PR #6, opened 2024-12-03
Add `mo:base/deprecated` directory for migration

rvanasa:


luc-blaeser (approved):
Good pragmatic approach. With that, we can also include migration functions inside the legacy data structures to transform to the corresponding new data structures.

rvanasa:
Converting to a draft PR for now and will update this to reflect changes to the original base library until we are ready for release.

luc-blaeser (approved):
I think this is a pragmatic approach of allowing gradual migration.

PR #47, opened 2024-12-13
Rename `insert()` functions

### LINKED ISSUE #43: legacy put vs new insert
So Andreas suggested the put/get naming, which is even in the guidelines I think, Is it worth renaming put to insert?

rvanasa:
Potentially resolves #43.

### DIFF
--- a/src/Map.mo
+++ b/src/Map.mo
@@ -37,7 +37,7 @@ module {
     todo()
   };
 
-  public func put<K, V>(map : Map<K, V>, key : K, value : V, compare : (K, K) -> Order.Order) : () {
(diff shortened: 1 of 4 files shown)


crusso (approved):
Let's see what @luc-blaeser says.

Swap might be a good name for the one that returns any previous binding...

rvanasa:
My hesitation with `Map.swap()` is that it could imply that the value is only added if another already exists with the same key. Definitely open to more opinions for this.

rvanasa:
Merging for now; feel free to open an issue for other naming suggestions.

PR #50, opened 2024-12-13
Experiment: `Iter.convert()`

rvanasa:
Idea for a convenience function to convert between data structures. The examples don't compile with the current Motoko version (hopefully there's a way around this), so the goal is to show the motivation for the `toIter()` and `fromIter()` module functions.

### DIFF
--- a/src/Iter.mo
+++ b/src/Iter.mo
@@ -72,7 +72,11 @@ module {
     todo()
   };
 
-  public func convert<A, B, T>(A : module { toIter : A -> Iter<T> }, B : module { fromIter : Iter<T> -> B }, a : A) : B {
+  public func convert<A, B, T>(
+    A : module { toIter : A -> Iter<T> },
+    B : module { fromIter : Iter<T> -> B },
+    a : A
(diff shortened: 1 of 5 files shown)


src/Iter.mo
```diff
@@ -71,4 +72,12 @@ module {
     todo()
   };
 
+  public func convert<A, B, T>(
+    A : module { toIter : A -> Iter<T> },
+    B : module { fromIter : Iter<T> -> B },
+    a : A
+  ) : B {
+    A.toIter(a) |> B.fromIter(_)
+  };
```
crusso, line 81: I think expressing this needs higher-kinded polymorphism:
```
  public func convert<A, B, T>(
    src : module { toIter : <T>A<T> -> Iter<T> },
    dst : module { fromIter : <T>Iter<T> -> B<T> },
    a : A<T>
  ) : B<T> {
    src.toIter(a) |> dst.fromIter(_)
  };
```

crusso: (Here, A and B are type constructors, not types)

PR #51, opened 2024-12-19
Update data structure APIs

rvanasa:
Changes in this PR:

* Rename `vals()` to `values()`, anticipating a corresponding `values()` alias in the Motoko compiler for Arrays, Blobs, and Text values
* Rename reverse iterator functions, e.g. `Map.reverseEntries()`
* Always include data structures as the first parameter in `Set.mo`, `Stack.mo`, etc. for consistency
* Change `flatten()` to take an iterator for data structure modules
* Remove `toIter()` in favor of explicit `values()`, `entries()`, etc.
* Map: remove `take()` and `insertTake()`
* Map: add `replaceIfExists()` and `filter()`
* Set: add `filter()`, `flatten()`, and `singleton()`
* Stack: swap order of args for `split()` and `chunks()`
* Vec: rename to `List`
* List: rename `isSubListOf()` to `containsAll()`
* List: remove `isStrictSubListOf()`, `chain()`, and `forNone()`
* List: add `flatMap()` to replace `chain()`
* List: rename `forAll()` to `all()`, `forEvery()` to `any()`, `dedupe()` to `distinct()`
* Array: Rename `init()` to `repeat()`


src/Array.mo
```diff
@@ -6,7 +6,9 @@ import { nyi = todo } "Debug";
 
 module {
   
-  public func init<T>(size : Nat, initValue : T) : [var T] = Prim.Array_init<T>(size, initValue);
+  public func new<T>() : [T] = [];
```
crusso, line 9: I'm slightly worried about using new (here and elsewere, e.g lists) in case we want to use it for classes at  some point. 
Does empty() work  as well? That would aslo work nicely with isEmpty() for structures that provide it.

crusso (approved):
Looks good. I worry that all/any is too easily misread compared with all/some, but not a biggie.

src/Array.mo
rvanasa: That's a good point about leaving this open for classes in the future. I'll create another PR for this.

PR #52, opened 2024-12-19
Add new Random API

rvanasa:
Because there is a wide design space for random number generation, it makes sense to defer most RNG functionality to external Mops packages for different use cases. However, it would be useful to have a basic set of operations and conventions which libraries could use as building blocks for more complex behavior. 

This PR proposes a minimal, extensible replacement to the `Random.Finite` class. We could add more functions over time for sufficiently common use cases.


src/Random.mo
```diff
@@ -1,15 +1,71 @@
...
+      todo()
+    };
+
+    public func natRange(min : Nat, maxExclusive : Nat) : async Nat {
+      todo()
+    };
+
+    public func intRange(min : Int, maxExclusive : Int) : async Int {
+      todo()
+    };
```
crusso, line 67: I wonder if something like this might be more efficient and useful, avoiding a full scheduler hop until more entropy is needed. Would need experimentation.

```suggestion
  public class AsyncRandom(generator : () -> async* Byte) {

    public func bool() : async* Bool {
      todo()
    };

    public func byte() : async* Nat8 {
      todo()
    };

    public func float() : async* Float {
      todo()
    };

    public func natRange(min : Nat, maxExclusive : Nat) : async* Nat {
      todo()
    };

    public func intRange(min : Int, maxExclusive : Int) : async* Int {
      todo()
    };
```

crusso (commented):
Would be good to get Timo's advice on this one...

src/Random.mo
rvanasa: I like that idea as long as you think it could realistically work in place of the shared function.

src/Iter.mo
```diff
@@ -8,6 +8,8 @@ module {
 
   public type Iter<T> = Type.Iter<T>;
 
+  public func empty<T>() : Iter<T> = { next = func() { null } };
```
ggreif, line 11:
```suggestion
  public func empty<T>() : Iter<T> = { next = func _ = null };
```
this should work

src/Random.mo
```diff
@@ -1,15 +1,87 @@
...
+    public func bool() : Bool {
+      todo()
+    };
+
+    /// Random `Nat8` value in the range [0, 256).
+    public func byte() : Nat8 {
+      todo()
+    };
+
+    /// Random `Float` value in the range [0, 1).
```
ggreif, line 48: maybe mention that the resolution is maximal (all relevant mantissa bits are random)

src/Random.mo
```diff
@@ -1,15 +1,87 @@
...
+              case (?n) n;
+              case null unreachable()
+            }
+          }
+        }
+      }
+    )
+  };
+
+  public class Random(generator : () -> Nat8) {
```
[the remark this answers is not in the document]
rvanasa: What do you think if we changed this to `generator : () -> Blob` as an optimization?

luc-blaeser (approved):
Nice, ergonomic design

PR #53, opened 2024-12-19
Rename `Debug.nyi()` to `Debug.todo()`

rvanasa:
From a readability standpoint, `todo()` is more likely to be understood at a glance compared to `nyi()`. This rename is also consistent with Rust in conjunction with the `Debug.unimplemented()` function.


crusso (commented):
Nice.
We could even add a special deprecation comment to warn about todos src/Debug.mo

crusso (approved):

PR #54, opened 2024-12-20
Replace `Iter.range()` with `Int.range()` and `Nat.range()`

### LINKED ISSUE #31: Exclusive `Iter.range()`?
Do we want to change the behavior of `Iter.range()` to exclude the upper bound? For example, `Iter.toArray(Iter.range(0, 3))` would return `[0, 1, 2]` instead of `[0, 1, 2, 3]`.

While an exclusive upper bound is the expected behavior for a zero-indexed language, making this change could silently break people's code during migration to the new base library. Opening this issue for a place to discuss whether we should do this.

rvanasa:
This PR removes the inclusive `Iter.range()` function in favor of an exclusive range function for each numeric type. In the future (or as part of this base library overhaul), we could also include range functions for fixed-size numeric types and perhaps also the `Char` module. 

Resolves #31.


src/Int.mo
```diff
@@ -58,4 +59,8 @@ module {
 
   public func pow(x : Int, y : Int) : Int { x ** y };
 
+  public class range(fromInclusive : Int, toExclusive : Int) {
```
crusso, line 62: Thinking ahead about to finite types, exclusive bounds won't let us specify a range up to the last element of a finite type).
Should we consider an API like

public class range(fromInclusive : T, {#Exc : T; #Inc : T) {..}

I guess the alternative is to provide 

public class rangeInc(fromInclusive : T, toExclusive : T) {..}


Also should the range count down if the bound is below the range? I don't know what we do at the moment.

rvanasa: That's a very good point. This would be my preference to simplify the most common case:

```motoko
public class range(fromInclusive : T, toExclusive : T) : T { ... }

public class rangeInclusive(from : T, to : T) { ... }
```

If it's potentially confusing for a finite type, we could rename `range()` to `rangeExclusive()` for that module to make it an explicit choice. 

We could also simplify the most likely use case by including `allValues()` functions for `Nat8`, `Int8`, `Bool`, `Compare`, etc.

rvanasa:
@crusso, is there a performance benefit to using classes in place of functions, or is this just for conciseness?

rvanasa:
Perhaps we could merge this for now and then make follow-up changes in a separate PR? Hoping to reduce the number of merge conflicts due to a lot of changes happening in parallel.

crusso (approved):
Looks good to me. Shall we add these to all other ordered types or leave for later?

rvanasa:
> Shall we add these to all other ordered types or leave for later?

I'll open a separate PR for the finite ordered types, since the design choices are a bit less clear with regards to `range()` and `rangeInclusive()`.

crusso (approved):

crusso:
> @crusso, is there a performance benefit to using classes in place of functions, or is this just for conciseness?
No, in fact using a class might be bit weird since it introduces a new type name (here lowercase) as well. I wonder why it was written as a class in the first place?

rvanasa:
Changed to use functions in the follow-up PR (https://github.com/dfinity/new-motoko-base/pull/86).

PR #56, opened 2024-12-20
Add `allValues()` functions to modules with finite types

rvanasa:
Provides a way to iterate through all possible values of finite types such as `Bool`, `Order`, `Nat8`, `Char`, etc.

Context: https://github.com/dfinity/new-motoko-base/pull/54#discussion_r1894380633.

### DIFF
--- a/src/Order.mo
+++ b/src/Order.mo
@@ -1,5 +1,8 @@
 /// Order
 
+import Iter "IterType";
+import { todo } "Debug";
+
 module {
 
   public type Order = {
@@ -23,5 +26,9 @@ module {
   public func equal(o1 : Order, o2 : Order) : Bool {
     o1 == o2
   };
+
+  public func allValues() : Iter.Iter<Order> {
+    todo()
+  }
 
 }
--- a/src/Bool.mo
+++ b/src/Bool.mo
@@ -1,6 +1,7 @@
 /// Boolean types and operations
 
 import Prim "mo:⛔";
+import Iter "IterType";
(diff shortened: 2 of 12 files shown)


src/Order.mo
```diff
@@ -24,4 +27,8 @@ module {
     o1 == o2
   };
 
+  public func allValues() : Iter.Iter<Order> {
```
crusso, line 30: Will anyone use this, I wonder?

crusso (commented):
[Naming a function to enumerate all values of an ordered type](https://claude.ai/chat/a4bb30f1-3659-4253-8465-33b7bd7553b7)



Claudio:
whats a better name than allValues() for a function that enumerates all values of an ordered type as an iterator?


Claude: Some better alternatives:

values() - Simple, standard in many languages
enumerate() - Clear if returning indexed values
range() - Good for numeric types
sequence() - Emphasizes ordering
values() is probably best as it's both concise and widely recognized. Do you need the function to have any special characteristics beyond basic enumeration?

...
iterate()
elements()
entries()
items()
getAll()
traversal()
stream()
...
series()
progression()
orderedValues()
valueIterator()
valueStream()
asIterator()
asSequence()

crusso (approved):

rvanasa:
(From a Slack conversation) `values()` could be confusing in the sense that there is also a function named `values()` on collection modules have a different purpose. `allValues()` was an adjustment from this to make it explicit that you're iterating through all possible values.

PR #86, opened 2025-01-07
Add validation system to ensure consistent functions across groups of modules

rvanasa:
This PR adds a validation script (available via `npm run validate`) which ensures that functions exist for a specified group of modules. Using this system, I added more functions across various modules for consistency. The config files are located in the `/validation` directory in the root of the repository ([example](https://github.com/dfinity/new-motoko-base/blob/spec/validation/collections.json)).

The validation config format is provisional, and we could turn into a more sophisticated system in the future if it ends up being useful beyond of the scope of the API redesign.


luc-blaeser (dismissed):
Nice tooling to keep our data structure APIs consistent.

luc-blaeser (approved):

PR #88, opened 2025-01-09
Rename `mo:base/pure/*` to `mo:base/functional/*`

rvanasa:
This PR renames the purely functional data structures from "pure" to "functional" as an alternative naming pattern. 

Pros:
* Clearly relates to functional programming.

Cons:
* Significantly longer to type.
* Potentially confusing for new developers who are familiar with functions but not functional programming.

Alternative to #89.

PR #89, opened 2025-01-09
Rename `mo:base/pure/*` to `mo:base/immutable/*`

rvanasa:
This PR renames the purely functional data structures from "pure" to "immutable" as an alternative naming pattern.

Rather than changing `toPure` and `fromPure` -> `toImmutable` and `fromImmutable`, I reverted to the current base library's convention of `freeze` and `thaw` in the `Array` module. 

Alternative to #88.


crusso:
I actually prefer `pure`, tbh

luc-blaeser (approved):
Thanks for this improvement!

PR #97, opened 2025-01-10
Add `VarArray` module

rvanasa:
This PR introduces an module with a similar interface to `Array` except with `[var T]` in place of `[T]` for args and return values.


crusso:
To be discussed, I think...

PR #106, opened 2025-01-15
Data structure API refinements

luc-blaeser:
Preparation for implementing `Map` and `Set` data structures:
1. Rename `pure` to `immutable` (merge PR https://github.com/dfinity/new-motoko-base/pull/89)
2. Put `compare` parameter in second place in `Map` and `Set` functions.


luc-blaeser:
Probably not needed if we can merge the other PR?

PR #107, opened 2025-01-15
Move `compare` function to second arg where relevant

rvanasa:
Moves the `compare` arg for all data structure operations with the exception of those comparing two data structures as the first two args. 

Supersedes #106.


luc-blaeser (approved):
Thanks for adjusting the APIs

PR #109, opened 2025-01-15
Add API validation in CI

rvanasa:
This PR adds a lockfile system which ensures that the PR does not modify the public API (`npm run validate` to override). This is intended as a sanity check for copy/pasting large amounts of code from the original base library. It currently uses a hacky regex parser, but we could eventually make this more sophisticated if it ends up being useful enough to keep post-release.

A secondary use case of this lockfile is that we can also generate one for the original base library and view the diff between the APIs as a resource for migration / documentation.


luc-blaeser (approved):
Nice. Thank you for this PR.

PR #110, opened 2025-01-15
Add implementations for original base library functions

rvanasa:
This PR includes adaptations of the existing function implementations and doc comments from the original base library.

**Progress:**

* [x] Doc comments
* [x] Implementations

I'm splitting the tests into a separate PR so that we can merge this more quickly and reduce the amount of merge conflicts for work happening in parallel. 

Depends on https://github.com/dfinity/new-motoko-base/pull/109 for API validation (see [`api.lock.json`](https://github.com/dfinity/new-motoko-base/blob/implement-existing/validation/api/api.lock.json)).


luc-blaeser (dismissed):
Wow, fantastic!
Maybe some follow-up PR: We could also copy/port tests for these original implementations.

luc-blaeser (approved):

PR #114, opened 2025-01-23
Implement `range()`, `rangeInclusive()`, and `allValues()`

### LINKED ISSUE #94: Implementation: `range()` and `allValues()` for numeric types
Size estimate: S

Implement, test, and document the new numeric functions such as `Nat8.range()` and `Nat8.allValues()`.

rvanasa:
Resolves #94.

The `range()` implementation in this PR is consistent with [Python's `range()` function](https://www.w3schools.com/python/ref_func_range.asp). When the first argument is greater than the second, the range functions return an empty iterator (same for Python, Rust, Java, and most other mainstream languages). This also applies for `rangeInclusive()`. Examples of each of these cases are included in the doc comments.

Alternatively, the `range()` function could count downwards if the first arg is greater than the second, although this might lead to unexpected behavior given that this is atypical for range implementations in mainstream languages. 

Once we decide on the implementation details, I will fill out the rest of the `Int*` and `Nat*` modules.

**Questions:**

* Do we want to include `rangeWithStep()` to pass a third interval argument, [similar to Python](https://docs.python.org/3/library/functions.html#func-range)? This would make it possible to express reverse ranges, e.g. `Int.rangeWithStep(4, 1, -1)`.
* For finite types, do we want to include `rangeWrap()` which wraps around from the max to min value?

We can address both of these later, but I wanted to mention these so we can plan ahead for future additions.

Unit tests pending https://github.com/dfinity/new-motoko-base/pull/116 (perhaps in a separate PR so we can review and merge these changes in parallel).


crusso (dismissed):

crusso:
LGTM sofar

rvanasa:
> Are these supposed to have entries in api.lock.json

The interface was already included in a previous PR, so this was just filling in the implementations.

PR #118, opened 2025-01-23
Replace `iterate()` functions

rvanasa:
Separates the `Iter.iterate()` function into two functions:

* `Iter.enumerate()` returns an iterator with indices.
* `Iter.forEach()` calls a function with each value and discards the return value.

This is a more flexible and composable design which is consistent with languages such as Python and Rust.

I also replaced `Option.iterate()` with `Option.forOk()` and `Option.forErr()` for naming consistency with `forEach()`, `mapOk()`, and `mapErr()`.


crusso (approved):
LGTM

PR #119, opened 2025-01-24
Implement `VarArray` functions

### LINKED ISSUE #98: Implementation: `VarArray` module
Size estimate: M

Implement, test, and document functions in the `VarArray` module proposed in #97.

rvanasa:
This PR implements most of the functions in the `VarArray` module along with making corresponding improvements to `Array` functions and documentation. 

An especially noticeable change in the diff is renaming `generate()` to `tabulate()`, which is rolling back to the name in the original base library. `tabulate()` seems preferable due to the naming of the `Array_tabulate` prim function and consistency with Scala's [`List.tabulate()`](https://www.educative.io/answers/what-is-tabulate-in-scala) pattern, but we might be able to come up with a better name for this function. 

Alternatives:

* `fromSize()`
* `build()`
* `make()`: already used in the original base library for constructing singleton arrays
* `new()`: naming might collide with future class-based constructors for other base library data structures

It's also worth mentioning that `Array.compare()` uses the same comparison logic as Python. If arrays have different sizes but all elements up to the shorter length are equal, the shorter array is considered less than the longer array.

Resolves #98.


src/Array.mo
```diff
@@ -522,43 +525,47 @@ module {
     acc
   };
 
-  /// Flattens the array of arrays into a single array. Retains the original
+  /// Flattens the iterator of arrays into a single array. Retains the original
```
crusso, line 528: I wonder if we should keep the original flatten, that can be achieved with a single allocation, and perhaps add an iterator  that can be implemented using that. 

The problem with this flatten is that you can't easily determine the size of the the array without iterating and storing the intermediate elements somewhere. Note you can't iterate twice to discover the total size and then again to append the elements,

src/Array.mo
```diff
@@ -21,33 +21,35 @@ import { todo } "Debug";
...
-  /// let array = Array.init<Nat>(4, 2);
+  /// let array = Array.repeat<Nat>("Echo", 3);
+  /// assert array == [var "Echo", "Echo", "Echo"];
   /// ```
   ///
   /// Runtime: O(size)
   ///
   /// Space: O(size)
-  public func init<T>(size : Nat, initValue : T) : [T] = Prim.Array_tabulate<T>(size, func _ = initValue);
+  public func repeat<T>(item : T, size : Nat) : [T] = Prim.Array_tabulate<T>(size, func _ = item);
```
crusso, line 37: For some reason, I'd expect the size to come first..

crusso: An alternative name might be `replicate` to rhyme with `tabulate`, but maybe repeat is established in other langs.

src/Array.mo
```diff
@@ -602,28 +609,62 @@ module {
   /// Space: O(1)
   public func values<T>(array : [T]) : Iter.Iter<T> = array.vals();
```
crusso, line 610: do we also want 
```
public func entries<T>(array : [T]) : Iter.Iter<(Nat, T)> = array.entries()
```

src/VarArray.mo
```diff
@@ -3,46 +3,162 @@
...
+  /// let array : [var Nat] = VarArray.tabulate<Nat>(4, func i = i * 2);
+  /// assert array == [var 0, 2, 4, 6];
+  /// ```
+  ///
+  /// Runtime: O(size)
+  ///
+  /// Space: O(size)
+  ///
+  /// *Runtime and space assumes that `generator` runs in O(1) time and space.
+  public func tabulate<T>(size : Nat, generator : Nat -> T) : [var T] {
```
crusso, line 67: Note to self: We should implement this as a runtime primitive, like Array_tabulate. It would be a lot more efficient (no bounds checks etc)

src/VarArray.mo
```diff
@@ -107,104 +223,740 @@ module {
...
+  ///     func(sumSoFar, x) = sumSoFar + x // this entire function can be replaced with `add`!
+  ///   );
+  /// ```
+  ///
+  /// Runtime: O(size)
+  ///
+  /// Space: O(1)
+  ///
+  /// *Runtime and space assumes that `combine` runs in O(1) time and space.
   public func foldLeft<T, A>(array : [var T], base : A, combine : (A, T) -> A) : A {
```
crusso, line 584: It's worth checking foldLeft and foldRight have consistent argument across all our collections.

src/VarArray.mo
```diff
@@ -107,104 +223,740 @@ module {
...
+  /// let array = [1, 9, 4, 8];
+  /// let bookTitle = VarArray.foldRight<Nat, Text>(array, "", func(x, acc) = toText(x) # acc);
+  /// ```
+  ///
+  /// Runtime: O(size)
+  ///
+  /// Space: O(1)
+  ///
+  /// *Runtime and space assumes that `combine` runs in O(1) time and space.
   public func foldRight<T, A>(array : [var T], base : A, combine : (T, A) -> A) : A {
```
crusso, line 608: See above

src/VarArray.mo
```diff
@@ -107,104 +223,740 @@ module {
...
+    let size = array.size();
+    var i = size;
+    while (i > 0) {
+      i -= 1;
+      acc := combine(array[i], acc)
+    };
+    acc
   };
 
+  /// Flattens the iterator of mutable arrays into a single mutable array. Retains the original
```
crusso, line 619: Again, I thin it would be good to keep the nested array version and then define one for iterators on top of that if possible.

src/VarArray.mo
```diff
@@ -107,104 +223,740 @@ module {
...
+  /// for (element in array.values()) {
+  ///   sum += element;
+  /// };
+  /// sum
+  /// ```
+  ///
+  /// Runtime: O(1)
+  ///
+  /// Space: O(1)
   public func values<T>(array : [var T]) : Iter.Iter<T> = array.vals();
```
crusso, line 701: Also implemet entries (as suggested above for immutable arrays)

src/VarArray.mo
```diff
@@ -107,104 +223,740 @@ module {
...
+  /// assert s.next() == null;
+  ///
+  /// let s = Array.slice<Nat>(array, 0, 0);
+  /// assert s.next() == null;
+  /// ```
+  ///
+  /// Runtime: O(1)
+  ///
+  /// Space: O(1)
   public func slice<T>(array : [var T], fromInclusive : Int, toExclusive : Int) : Iter.Iter<T> {
```
crusso, line 861: I wonder how useful this notion of slice is, since its an iterator.
If this is new, maybe we should postpone its introduction until we decide what a useful notion of slice is.
Maybe we actually want a view on the array with random, not sequential, access to elements.

crusso:
Added some comments. Not sure if they conflict with other advice you got so feel free to push back.

src/VarArray.mo
crusso: Also why are the bounds Int and not Nat?

src/Array.mo
rvanasa: That's a good point. Maybe we could rename this to `flattenArray()` so that `flatten()` has a consistent API across different modules. I'll go ahead with this for now, and then we could revisit if it seems better for it to be the other way around.

src/Array.mo
rvanasa: `repeat()` comes from Java and JS strings, where I've extended this naming convention to other modules for consistency.

src/Array.mo
rvanasa: Good idea; I think we could reasonably do that. Do we want to include `entries()` in the compiler? Otherwise, we could maybe just implement this in the module itself.

Also, I wonder if it would make sense to call this `Array.enumerate()` for consistency with `Iter.enumerate()`. The precedent from other languages is that `entries()` applies when you have explicit key-value pairs whereas `enumerate()` has implicit indices for an ordered collection, although this situation is a bit of a grey area.

src/VarArray.mo
rvanasa: Yes this would be good, since it's probably the most frequently-used function in this module for the base library implementation. Perhaps `VarArray_tabulate()`?

src/VarArray.mo
rvanasa: Looks like they're currently all consistent. That's a good point knowing that some languages and libraries will swap the arguments for `foldRight()`, which I'd prefer to avoid.

src/VarArray.mo
rvanasa: That's a really good point about potentially wanting to return a view rather than an iterator. Maybe we could rename `slice()` to `range()` due to working in a similar way to ranges over fixed-size numeric types (except where negative numbers are relative to the end of the array)? I updated the PR to show what this would look like, but we could still remove it if you think this might be confusing.

crusso (commented):
All looks good but I'll wait for the mods discussed on slack.

src/Array.mo
crusso: As in let's stick with this.

crusso: Though I guess there is some risk that people will rename from init without swapping argument order...

src/Array.mo
```diff
@@ -602,28 +645,91 @@ module {
...
   /// ```motoko include=import
   ///
-  /// let array = [1,2,3,4,5];
+  /// let array = [1, 2, 3, 4, 5];
   /// let subArray = Array.subArray<Nat>(array, 2, 3);
   /// ```
   /// Runtime: O(length)
   ///
   /// Space: O(length)
   public func subArray<T>(array : [T], start : Nat, length : Nat) : [T] {
```
crusso, line 731: I wonder if we want this to have the same parameters as range (or maybe  variant that does).
I guess it depends on what other languages do...

crusso (approved):
LGTM (feel free to merge ignoring the question about subArray). Maybe rangeArray could be added....

src/Array.mo
rvanasa: Most languages use a start/end index for this function instead of the length of the array. Because this could create breaking changes if we change the parameters, we might want to rename the function if we do that. I'll try a few different possibilities and open a PR if something seems to work nicely.

crusso (approved):

PR #121, opened 2025-01-28
Miscellaneous

rvanasa:
* Requires `<system>` capability to call `CertifiedData.set()`.
* Updates `Timer` and `Error` modules to reflect changes to original base library. 
* Removes deprecated hash functions.
* Updates trap messages to include the name of the relevant base library function.


src/Array.mo
```diff
@@ -397,7 +397,7 @@ module {
                   element
                 };
                 case null {
-                  Prim.trap "Malformed array in mapResults"
+                  Prim.trap "Array.mapResult(): Malformed array"
```
crusso, line 400:
```suggestion
                  Prim.trap "Array.mapResult(): malformed array"
```
Or make the one one on 335 consistent with this?

src/CertifiedData.mo
```diff
@@ -32,7 +32,9 @@ module {
   ///
   /// See a full example on how to use certified variables here: https://github.com/dfinity/examples/tree/master/motoko/cert-var
   ///
-  public let set : (data : Blob) -> () = Prim.setCertifiedData;
+  public func set<system>(data : Blob) : () {
+    Prim.setCertifiedData(data)
```
crusso, line 36: I guess we may need to make the Prim.setCertifiedData take a <system> argument too to make this really safe (otherwise people can directly call Prim.setCertifiedData).

src/Float.mo
```diff
@@ -507,25 +499,25 @@ module {
...
   /// Example:
   /// ```motoko
   /// import Float "mo:base/Float";
   ///
   /// let epsilon = 1e-6;
-  /// Float.equalWithin(-12.3, -1.23e1, epsilon) // => true
+  /// Float.equal(-12.3, -1.23e1, epsilon) // => true
   /// ```
-  public func equalWithin(x : Float, y : Float, epsilon : Float) : Bool {
+  public func equal(x : Float, y : Float, epsilon : Float) : Bool {
```
crusso, line 517: @luc-blaeser are you ok with renaming equalWIthin (and notEqualWithin) to equal and notEqual?
Also another option might be to make both of these curried functions, that take the epsilon as a first argument and return an equality/inequality function of the usual binary type:

```
  public func equalWithin(epsilon: Float) : (x : Float, y : Float) - > Bool =
       func(x : Float, y: float)  { ... epsilon ...  }
  }
 (and similarly for nonEqualWithin 
```

src/Hash.mo
```diff
@@ -22,29 +22,4 @@ module {
     ha == hb
   };
 
-  /// Computes a hash from the least significant 32-bits of `n`, ignoring other bits.
```
crusso, line 25: If we aren't going to provide and hash based datastructures (are we?) I'd be tempted to get rid of this whole module Hash. length seemed misnamed (should be maxBitIndex or something) and `bit` is probably an implementation detail of Trie.mo which uses the hash bits to index into a radix tree (IIRC). Might be worth checking if pos/length are used anywhere else.

crusso (approved):
LGTM (see suggestions)

src/CertifiedData.mo
rvanasa: > I guess we may need to make the Prim.setCertifiedData take a argument too to make this really safe

This is probably worth doing, since otherwise there isn't much of a benefit to changing the base library if people can go around it by importing `Prim` directly. 

I'll roll this back to `public let set : (data : Blob) -> () = Prim.setCertifiedData` so that it passes on whether the prim function has the `<system>` requirement.

src/Float.mo
rvanasa: The motivation here is for API uniformity. I suppose we could curry the arguments, although I think `equal(x : Float, y : Float, epsilon : Float)` is more consistent with how we handle the `compare` and `equal` functions for data structures.

src/Hash.mo
rvanasa: I'll open a PR for this, since this seems reasonable.

src/Float.mo
crusso: Right, I thinking about the situation where you want to pass the equality function as a first-class argument (which is why we define it for convenience...)

I don't have a strong opinion...

src/Float.mo
```diff
@@ -538,23 +530,27 @@ module {
...
   /// ```motoko
   /// import Float "mo:base/Float";
   ///
   /// let epsilon = 1e-6;
-  /// Float.notEqualWithin(-12.3, -1.23e1, epsilon) // => false
+  /// Float.notEqual(-12.3, -1.23e1, epsilon) // => false
   /// ```
-  public func notEqualWithin(x : Float, y : Float, epsilon : Float) : Bool {
-    not equalWithin(x, y, epsilon)
+  public func notEqual(x : Float, y : Float, epsilon : Float) : Bool {
```
crusso, line 548: Do we even want notEqual, since its the literal negation of equals @luc-blaeser ?

rvanasa: Seems reasonable to remove this function. I'll merge and open another PR to start working on the merge conflicts in parallel.

PR #124, opened 2025-01-29
Add `step()` function (`range()` with a custom step size)

rvanasa:
This PR includes a function which iterates over a range of numbers with a given positive or negative step size with the same behavior as Python's `range()` function with three arguments. 

`Nat.range(a, b)` is equivalent to `Nat.step(a, b, 1)`. 

Alternative names:

* `Nat.steps()`
* `Nat.rangeWithStep()`
* `Nat.rangeStep()`

In addition to numeric types, I also included `Iter.step()` which returns every nth item from an iterator.

If we decide to include this, I will fill out the rest of the `Nat*` and `Int*` implementations before merging.

Unit tests pending #116.


validation/api/api.lock.json
```diff
@@ -467,18 +467,19 @@
...
-      "public func enumerate<T>(xs : Iter<T>) : Iter<(Nat, T)>",
-      "public func filter<T>(xs : Iter<T>, f : T -> Bool) : Iter<T>",
-      "public func forEach<T>( xs : Iter<T>, f : (T) -> () )",
-      "public func fromArray<T>(xs : [T]) : Iter<T>",
-      "public func fromVarArray<T>(xs : [var T]) : Iter<T>",
+      "public func enumerate<T>(iter : Iter<T>) : Iter<(Nat, T)>",
+      "public func filter<T>(iter : Iter<T>, f : T -> Bool) : Iter<T>",
+      "public func forEach<T>( iter : Iter<T>, f : (T) -> () )",
+      "public func fromArray<T>(iter : [T]) : Iter<T>",
+      "public func fromVarArray<T>(iter : [var T]) : Iter<T>",
```
crusso, line 474:
```suggestion
      "public func fromArray<T>(array : [T]) : Iter<T>",
      "public func fromVarArray<T>(array : [var T]) : Iter<T>",
```

src/Iter.mo
```diff
@@ -202,16 +233,16 @@ module {
   /// assert(?3 == iter.next());
   /// assert(null == iter.next());
   /// ```
-  public func fromArray<T>(xs : [T]) : Iter<T> {
-    var ix : Nat = 0;
-    let size = xs.size();
+  public func fromArray<T>(iter : [T]) : Iter<T> {
```
crusso, line 236:
```suggestion
  public func fromArray<T>(array : [T]) : Iter<T> {
```

src/Iter.mo
```diff
@@ -221,8 +252,8 @@ module {
   /// Like `fromArray` but for Arrays with mutable elements. Captures
   /// the elements of the Array at the time the iterator is created, so
   /// further modifications won't be reflected in the iterator.
-  public func fromVarArray<T>(xs : [var T]) : Iter<T> {
-    fromArray<T>(Array.fromVarArray<T>(xs))
+  public func fromVarArray<T>(iter : [var T]) : Iter<T> {
```
crusso, line 255:
```suggestion
  public func fromVarArray<T>(array : [var T]) : Iter<T> {
```

src/Iter.mo
```diff
@@ -221,8 +252,8 @@ module {
...
+    fromArray<T>(Array.fromVarArray<T>(iter))
```
crusso, line 256:
```suggestion
    fromArray<T>(Array.fromVarArray<T>(array))
```
Is copying the array intentional? To provide a safe snapshot, I guess?

rvanasa: I think I just did this for conciseness, but it definitely seems worth the repeated code to remove the performance overhead of `fromVarArray()`.

validation/api/api.lock.json
rvanasa: Thanks for catching this; updated.

src/Nat.mo
```diff
@@ -386,6 +386,63 @@ module {
...
+  /// assert(?7 == iter2.next());
+  /// assert(?5 == iter2.next());
+  /// assert(?3 == iter2.next());
+  /// assert(null == iter2.next());
+  /// ```
+  ///
+  /// If `step` is 0 or if the iteration would not progress towards the bound, returns an empty iterator.
+  public func step(fromInclusive : Nat, toExclusive : Nat, step : Int) : Iter.Iter<Nat> {
+    if (step == 0) {
+      Iter.empty()
```
crusso, line 412: I think Python actually traps on step == 0, if we are trying to match Python. Not sure what is does about lack of progress.

crusso:
Maybe having a name prefixed with range would be better for intellisense. Another option might be `rangeBy` although rangeWithStep is clear too.

src/Iter.mo
```diff
@@ -216,16 +248,16 @@ module {
   /// assert(?3 == iter.next());
   /// assert(null == iter.next());
   /// ```
-  public func fromArray<T>(xs : [T]) : Iter<T> {
-    var ix : Nat = 0;
-    let size = xs.size();
+  public func fromArray<T>(array : [T]) : Iter<T> {
```
crusso, line 251: I'm not sure we should worry about but both fromArray and fromVarArray have (and always have had) a small space leak, because the iterator will hold on to the source array even after iteration has finished. @luc-blaeser is this worth fixing or should we just live with it?

crusso (approved):
But see suggestions.

src/Nat.mo
rvanasa: That's correct; Python raises an error when step is 0. 

I suppose we could remain fully consistent, although it seems a bit odd to trap for this but not when the lower bound is greater than the upper bound. Maybe we could trap for both conditions to reduce the chance of unintended bugs, even though this is a bit stricter than the behavior in Python?

rvanasa:
> Maybe having a name prefixed with range would be better for intellisense. Another option might be rangeBy although rangeWithStep is clear too.

`rangeWithStep()` is probably still my preference. @ggreif, does the intellisense point change your opinion, or would you still prefer `steps()` / `step()`?

Issue #95, opened 2025-01-10
Implementation: type conversions


rvanasa:
Size estimate: S

Add missing implementations, tests, and docs for type conversions (especially for data structures). Most of these new conversions should be trivial to implement.

PR #125, opened 2025-01-29
Add `Int` / `Nat` conversions

rvanasa:
This PR introduces `Int.toNat()`, `Int.fromNat()`, `Nat.toInt()`, and `Nat.fromInt()` to convert between integers and natural numbers. Because this is such a common use case, I think these conversions are worth including in both modules for convenience. 

In addition, I added `Int.fromText()` which works in the same way as `Nat.fromText()` with an optional `+` or `-` symbol at the beginning of the input text. 

Progress for #95.


crusso (dismissed):
LGTM!

crusso (approved):
Sorry for the delay, didn't see the request. `fromText` i s another operation we might want to move to the RTS for speed... but can definitely come later...

PR #126, opened 2025-01-31
Add `sliceToArray()` in `Array` and `VarArray`

rvanasa:
Follow-up to address [this comment](https://github.com/dfinity/new-motoko-base/pull/119#pullrequestreview-2584681089) on #119.

The naming depends on whether we eventually want to include a `slice()` function which returns a view (rather than copy) of the original array. If not, another design consideration is whether we want `VarArray.slice()` to return `[T]` or `[var T]`. This is the reasoning for the current name (`sliceToArray()`), which would give us the flexibility to also include `VarArray.sliceToVarArray<T>(...) : [var T]` in the future.


luc-blaeser (dismissed):
Nice!

crusso (approved):

PR #129, opened 2025-02-04
Remove hashing

rvanasa:
This PR removes the default hash functions from the base library in favor of using Mops packages within the context of a particular use case. 

With the new base library, we are moving away from hash-based data structures due to the ability for attackers to exploit hash collisions to significantly degrade canister performance. All built-in data structures are now using comparison rather than hashing, and it could make sense to further encourage this best practice by removing hashing from the base library. 

Follow-up for https://github.com/dfinity/new-motoko-base/pull/121#discussion_r1941149659.


crusso (commented):
My original intentions was just to remove Hash.mo, but maybe this more drastic change is fine too.

I expect some people may complain in which case we can put some back selectively.

crusso (approved):

rvanasa:
Closing in favor of #133.

PR #130, opened 2025-02-05
Refactor to use `Types` module

rvanasa:
Replaces `types/Iter.mo` with a `Types` module which contains all base library types. This (possibly along with a `Convert` module) will be important for defining type conversions without introducing circular dependencies.

Progress towards #95.


src/Types.mo
```diff
@@ -0,0 +1,94 @@
...
+
+  public type Hash = Nat32;
+  public type Iter<T> = { next : () -> ?T };
+  public type Order = { #less; #equal; #greater };
+  public type Result<T, E> = { #ok : T; #err : E };
+  public type Pattern = {
+    #char : Char;
+    #text : Text;
+    #predicate : (Char -> Bool)
+  };
```
crusso, line 32: i wonder it's still better to keep that in Text. Ditto for the other data structure ones unless really necessary.

src/Types.mo
```diff
@@ -0,0 +1,94 @@
...
+  public type Nat = Prim.Types.Nat;
+  public type Nat8 = Prim.Types.Nat8;
+  public type Nat16 = Prim.Types.Nat16;
+  public type Nat32 = Prim.Types.Nat32;
+  public type Nat64 = Prim.Types.Nat64;
+  public type Principal = Prim.Types.Principal;
+  public type Region = Prim.Types.Region;
+  public type Text = Prim.Types.Text;
+
+  public type Hash = Nat32;
```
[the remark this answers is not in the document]
rvanasa: I'll remove this as part of #129 if we end up merging that PR.

src/Types.mo
rvanasa: > Ditto for the other data structure ones unless really necessary.

I suppose we could move the other other data structure ones such as `Map.Node`, although in that case it's used in `Map`, so we would otherwise need to redefine it in both places. Makes it a somewhat tricky design choice. 

The vision for this module is that it would be possible to access all types defined in the base library, i.e. it might be odd to include every type except for a few select ones such as `Pattern`. I'd also be fine with removing it and then reintroducing later based on feedback.

crusso (approved):

rvanasa:
Merging for now, and we will revisit before releasing the new base library.

PR #100, opened 2025-01-13
Imperative btree

ByronBecker:
Sorry for the delay on this @rvanasa 😅 

I copied over all the files from the BTree, putting the main (exposed file) at BTree/lib.mo so that mo:base/BTree imports work.

I also copied over the tests, but feel free to rearrange in terms of how you want to run/organize those tests (doesn't seem like that's decided yet).

Let me know if you have any questions, and feel free to massage the contents of the PR to your liking!

PR #134, opened 2025-02-06
Set implementation

luc-blaeser:
Based on Byron Becker's imperative B-tree.
For performance, adjusted to ordered elements instead of key-value pairs.

Source: https://github.com/canscale/StableHeapBTreeMap
See PR: https://github.com/dfinity/new-motoko-base/pull/100

Implementation of `Set`, including:
* Unit tests
* Large-scaled randomized tests
* API documentation

* Implement conversion to and from immutable/Set.


src/immutable/Set.mo
```diff
@@ -103,7 +103,7 @@ module {
   public func foldRight<T, A>(
     set : Set<T>,
     base : A,
-    combine : (A, T) -> A
+    combine : (T, A) -> A
```
rvanasa, line 106: @crusso, what are your thoughts on the order of args here (since we briefly talked about this at one point)?

rvanasa (dismissed):
Fantastic; thank you! LGTM from an initial review. 

We could fix the CI errors by adding placeholder functions or relaxing the constraints in `/validation/collections.json`. 

It might also be necessary to regenerate the API lockfile with `npm run validate`, which confirms intentional changes to the public API.

luc-blaeser:
Thanks, Ryan, for looking at PR. Just fixed CI and merged latest.
We still need to implement and test conversion to/from immutable twin data structures.

src/immutable/Set.mo
crusso: I'd also got for consistency. Both Haskell and OCaml put the the accumulator on the right in foldRight and left in foldLeft. 


Caml even switches the order of the three arguments to match (but Haskell does not):

https://hackage.haskell.org/package/base-4.21.0.0/docs/Data-List.html#g:3

foldl :: [Foldable](https://hackage.haskell.org/package/base-4.21.0.0/docs/Data-Foldable.html#t:Foldable) t => (b -> a -> b) -> b -> t a -> b

foldr :: [Foldable](https://hackage.haskell.org/package/base-4.21.0.0/docs/Data-Foldable.html#t:Foldable) t => (a -> b -> b) -> b -> t a -> b

Ocaml: https://ocaml.org/manual/5.2/api/List.html

val fold_left : ('acc -> 'a -> 'acc) -> 'acc -> 'a list -> 'acc
fold_left f init [b1; ...; bn] is f (... (f (f init b1) b2) ...) bn.

val fold_right : ('a -> 'acc -> 'acc) -> 'a list -> 'acc -> 'acc
fold_right f [a1; ...; an] init is f a1 (f a2 (... (f an init) ...)). Not tail-recursive.

crusso: I must admit the Ocaml one seems the most principled.

crusso: Our own OrderedMap and Set use the same signature for both combines. which was perhaps a mistake.

rvanasa (approved):
LGTM and thanks again. I'll merge this now, and we can make follow-up changes in separate PRs.

PR #136, opened 2025-02-07
port OrderedMap (WIP)

crusso:
Port of Serokell's OrderedMap with explicit comparison functions (grrr).
- matching imperative API (pending #150)
- includes revamped doc (not yet checked with interpreter)
- tested

Remaining issues (can be resolved after merge):
#155
#154


src/Map.mo
```diff
@@ -1172,7 +1172,7 @@ module {
   ///   Map.add(map, Nat.compare, 2, "Two");
   ///
   ///   let text = Map.toText<Nat, Text>(map, Nat.toText, func (value) { value });
-  ///   // `(0, Zero), (1, One), (2, Two)`
+  ///   // `{(0, Zero), (1, One), (2, Two)}`
```
rvanasa, line 1175: @crusso @lucblaeser, maybe we could use a more syntax-like text representation (similar to JS and Python):

```suggestion
  ///   // `{ 0 = Zero; 1 = One; 2 = Two }`
```

rvanasa, line 1175: @crusso @lucblaeser, maybe we could use a more syntax-like text representation (similar to JS and Python):

```suggestion
  ///   // `{ 0 = Zero; 1 = One; 2 = Two }`
```

src/Map.mo
```diff
@@ -1183,13 +1183,13 @@ module {
   ///
   /// Note: Creates `O(log(n))` temporary objects that will be collected as garbage.
   public func toText<K, V>(map : Map<K, V>, keyFormat : K -> Text, valueFormat : V -> Text) : Text {
-    var text = "";
+    var text = "{";
```
rvanasa, line 1186: (See above comment)

```suggestion
    if (size(map) == 0) {
      return "{}"
    };
    var text = "{ ";
```

src/Map.mo
```diff
@@ -1183,13 +1183,13 @@ module {
...
+    var sep = "";
     for ((key, value) in entries(map)) {
-      if (text != "") {
-        text #= ", "
-      };
-      text #= "(" # keyFormat(key) # ", " # valueFormat(value) # ")"
+      text #= sep # "(" # keyFormat(key) # ", " # valueFormat(value) # ")";
```
rvanasa, line 1189:
```suggestion
      text #= sep # keyFormat(key) # " = " # valueFormat(value);
```

src/Map.mo
```diff
@@ -1183,13 +1183,13 @@ module {
...
+      sep := ", "
     };
+    text #= "}";
```
rvanasa, line 1192:
```suggestion
    text #= " }";
```

src/Types.mo
```diff
@@ -86,9 +86,21 @@ module {
...
+    public type Tree<K, V> = {
+      #red : (Tree<K, V>, K, V, Tree<K, V>);
+      #black : (Tree<K, V>, K, V, Tree<K, V>);
+      #leaf
+    };
+
     public type Queue<T> = (Stack.Stack<T>, Stack.Stack<T>);
     public type Set<T> = (); // Placeholder
-    public type Stack<T> = ?(Stack<T>, T)
+    public type Stack<T> = ?(Stack<T>, T); // Delete or reorder me
```
rvanasa, line 103: We can remove this:

```suggestion
```

rvanasa (commented):
Thanks! LGTM so far. I added a suggestion for `Map.toText()` and skimmed the immutable Map implementation.

src/Map.mo
crusso: That would be easier to read, but harder to cut-n-paste into a program, i.e. as the interior of an array. Open to suggestions.
I essentially used the relation representation of map as a set of pairs.

test/Map.test.mo
```diff
@@ -594,7 +594,7 @@ run(
           let map = Map.singleton<Nat, Text>(1, "1");
           Map.toText<Nat, Text>(map, Nat.toText, func(value) { value })
         },
-        M.equals(T.text("(1, 1)"))
+        M.equals(T.text("{(1, 1)}"))
```
rvanasa, line 597: This is what people coming from JS / Python would expect from the `toText()` function:

```suggestion
        M.equals(T.text("{ 1 = 1 }"))
```

rvanasa (approved):
Thank you for taking this on!

test/Map.test.mo
crusso: Right, but is that because it's valid syntax for constructing  map in JS/Python?

crusso: Ideally, I'd like to be able to cut-n-paste the toText output into some Motoko code to reconstruct the value, but perhaps that's just unrealistic.

src/pure/Map.mo
```diff
@@ -1,116 +1,1544 @@
...
     map : Map<K, V>,
     base : A,
     combine : (A, K, V) -> A
-  ) : A {
-    todo()
-  };
+  ) : A
+    = Internal.foldLeft(map.root, base, combine);
 
+  // TODO: base last?
```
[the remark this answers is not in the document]
crusso: Opened issue https://github.com/dfinity/new-motoko-base/issues/154

src/Map.mo
crusso: Opened issue #155

crusso:
Ryan, i've opened issues for the toText and foldRight questions, so I think we could merge this for now and revisit.

src/Map.mo
rvanasa: I believe so, since it's consistent with Motoko's code style. 

Since the `toText()` format is highly subjective, a resemblance to the Motoko language syntax is probably the least opinionated choice (essentially a [focal point](https://en.wikipedia.org/wiki/Focal_point_(game_theory))).

rvanasa (approved):

PR #137, opened 2025-02-07
Update immutable Stack representation

rvanasa:
Adjusts the `immutable/Stack` representation for compatibility with the original base library's `Stack` and `List` data structures.


src/Types.mo
```diff
@@ -89,6 +89,6 @@ module {
     public type Map<K, V> = (); // Placeholder
     public type Queue<T> = (Stack.Stack<T>, Stack.Stack<T>);
```
crusso, line 90: If the intention is to keep the old functional Queue from base, then this should be:
```suggestion
    public type Queue<T> = (Stack<T>, Stack<T>);
```
(otherwise I think it's using the imperative stack.

crusso (approved):
But see suggestion about Queue...

src/Types.mo
rvanasa: Oh that's a good point; this was a copy/paste error. Thanks for catching this.

rvanasa:
Since the PR got unapproved, I'll wait to merge this until we decide on the new names for `Stack` and `List`.

crusso:
I think we decided to name immutable/Stack to pure/List

rvanasa:
Closing; see #138.

PR #138, opened 2025-02-10
Rename `immutable` to `pure` and `pure/Stack` to `pure/List`

rvanasa:
Follow-up from an internal conversation. 

This PR makes the following renaming changes:

* `mo:base/immutable/*` -> `mo:base/pure/*`
* `mo:base/pure/Stack` -> `mo:base/pure/List`
* `freeze()` -> `toPure()`
* `thaw()` -> `fromPure()`


crusso (approved):
Let's go for this for now. I admit, it does read a bit funny, but shorter than `persistent` I guess.

PR #139, opened 2025-02-10
Implement `Array.fromIter()` and `VarArray.fromIter()`

rvanasa:
Implements `fromIter()` for `Array` and `VarArray` and enables tests which rely on these functions. This PR also includes some other tangential changes such as updating the compiler to the latest version. 

Note that this PR is set to auto-merge on approval.


crusso (commented):
LGTM.

Eventually, we can re-implement fromIter using Timo's Vector lib which should be more efficient than using List.

crusso (approved):

PR #140, opened 2025-02-10
Add `rangeBy()` and `rangeByInclusive()` to `Nat` and `Int` modules

rvanasa:
Continuation of #124.

* [x] Rename `step()` to `rangeBy()` (we could perhaps use another function name)
* [x] Implement `rangeByInclusive()` for `Nat` and `Int`

We can add the fixed-size numeric modules in a future PR, since there is some ambiguity about which data type to use for the fixed-sized `Nat*` step size.


luc-blaeser (approved):
Thank you!

PR #150, opened 2025-02-12
fix: imperative arrays: fix cloning, re-implement map, optimize uses of tabulate

### LINKED ISSUE #146: (Imperative) Map.map takes a redundant comparison function
This fails to match the signature for pure/Map.map, which doesn't not require a comparison because it copies the structure of the tree instead of reconstructing it from scratch like (imperative) Map.map which could do something similar I suspect.

crusso:
Map.clone was failing to clone leaf nodes, leading to (very likely) accidental sharing

Re-implement Map.map to make a copy with map, rather than rebuilding the map, removing the need for passing in a comparsion function.

Optimize uses of VarArray.tabulate to repeat.

Fixes #146 

Also hide internal types.

- added more tests of cloning


src/Types.mo
```diff
@@ -61,7 +61,7 @@ module {
       children : [var ?Node<K, V>]
     };
 
-    public type Leaf<K, V> = {
+    public type Leaf<K, V> = { // why the extra indirection?
```
rvanasa, line 64: Seems like we could remove `Leaf<K, V>` and then rename `Data` to `Leaf` for clarity?

rvanasa (approved):
LGTM, thanks!

src/Types.mo
crusso: Yeah, but perhaps Luc was intending to keep compatibility with the original library. If we do this, I would keep the Data name and just remove the Leaf type (since internal nodes also carry Data).

crusso: @luc-blaeser are we free to simplify the representation or do we want to preserve compatibility with Byron's representation for easier migration?

PR #159, opened 2025-02-14
feat: implement `put` `update` `swap` for Map and pure/Map

### LINKED ISSUE #156: Map.add and pure Map.add have inconsistent semantics
The former traps on override of an existing key, the latter does not (which is what I would expect from a map). 
I can make the semantics of `pure/Map.add` match `Map.add`, but I suggest we also support variants of add that 1) never trap (adding an entry if necessary) and 2) trap on absence of an entry.

How about the following:

Imperative:

```
add<K,V>(c, k, v) : () // traps on override of k
put<K,V>(c, k, v) : () // adds or updates existing k, never traps
update<K, V>(c, k, v) : () // updates k, traps on absent k.
```

And we rename the current `put` (corresponding to OrderedMap.replace, IIRC) to 

```
swap<K,V>(c, k, v) : ?K
```

Functional:

```
add<K,V>(c, k, v)  : C<K,V>  traps on override of k
put<K,V>(c, k, v)   : C<K,V>  adds or updates k with never traps
update<K, V>(c, k, v) : C<K,V> updates k, traps on absent k.
swap<K,V>(c, k, v) : (C<K,V>, ?K)  // : (?K, C<K,V>)
```

Pros: 

* `put` has the same semantics as base/OrderedMap.put,
* 
* `swap` is shorten and more descriptive than base/OrderedMap.replace or  the new-base/Map.put (that puts and returns the old value).

* `exchange`/`replace` are possible alternatives for `swap`

Cons:

Maybe other PLs have different conventions for this, but we can't be consistent with everything out there.

crusso:
Fixes #156 

map.add traps on re-entry
map.put adds or updates an entry
map.update updates an existing entry
map.swap puts and returns previous value, if any.


src/pure/Map.mo
```diff
@@ -179,8 +178,8 @@ module {
   public func get<K, V>(map : Map<K, V>, compare : (K, K) -> Order.Order, key : K) : ?V
     = Internal.get(map.root, compare, key);
 
-  /// Given `map` ordered by `compare`, add a mapping from `key` to `value`. Overwrites any existing entry with key `key`.
-  /// Returns a modified map.
+  /// Given `map` ordered by `compare`, add a new mapping from `key` to `value`.  /// Traps if `map` contains any existing entry for `key`.
```
crusso, line 181:
```suggestion
  /// Given `map` ordered by `compare`, add a new mapping from `key` to `value`.
  /// Traps if `map` contains any existing entry for `key`.
```

luc-blaeser (approved):
Thanks for harmonizing the API.

PR #165, opened 2025-02-19
Add tests for `VarArray`

rvanasa:
Adapts existing tests from `Array.test.mo` to cover the `VarArray` module.

### DIFF
--- a/src/VarArray.mo
+++ b/src/VarArray.mo
@@ -78,6 +78,39 @@ module {
     array
   };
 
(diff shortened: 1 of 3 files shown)


crusso (approved):
Very nice. I'll admit, I did not read all the tests. Let me know if you actually expected a detailed review of those.

rvanasa:
The tests are just a copy/paste of `Array.test.mo` with systematic adjustments such as replacing `[...]` with `[var ...]`, so that should be ok.

Issue #160, opened 2025-02-14
Map.delete and pure/Map.delete have inconsistent semantics


crusso:
The former traps deleting an absent key, the latter does not.

I suggest 

add/remove trapping
put/delete non-trapping

rvanasa:
What about something along the lines of `add() : ()` and `remove() : ()` (non-trapping) vs. `addSafe() : Bool` and `removeSafe() : Bool` (also non-trapping)? We could eventually include other variations such as `addChecked()` and `removeChecked()` (trapping), although I think it would be preferable to avoid trapping in favor of giving the developer a way to handle the error in context. 

It might be good to keep "add" and "remove" (or I suppose "put" and "delete") in both of the names if possible, since otherwise it could be confusing to new developers who might mix them up.

crusso:
Boolean add and remove would solve the problem but might lead to awkward code with all those ignores in the common case. And would 'addAll' etc return a bool and if so, with what semantics ? 

Maybe just non-trapping add/remove
with boolean put/delete?

I was also wondering if we could use

Union/intersection/difference (creating a new map)
And
Unite/intersect/diff for inplace update (in the imperative API)

Let's talk tomorrow...

rvanasa:
> Maybe just non-trapping add/remove with boolean put/delete?

I'd be ok with this, although a remaining concern is that it might be confusing to remember which ones return a boolean.

> Unite/intersect/diff for inplace update (in the imperative API)

In this case, we might want to rename the functions `sort()` -> `sorted()` and `sortInPlace()` -> `sort()`. Otherwise, I suppose `unionInPlace()`, `intersectInPlace()`, `diffInPlace()`, etc. could also work as a (much more verbose) naming convention. 

Overall, I've been choosing names based on consistency first, clarity second, and conciseness third as an explanation for thinking `*InPlace()` could make sense despite being verbose. This also comes from the Python ecosystem where data structures (primarily dataframes) have in-place operations such as `df.sort(..., inplace=True)`.

crusso:
The other thing that strikes me is that the Boolean ones are truly awkward in the functional api since you need to return a new map and a Boolean. Wouldn't be so bad if we had subtyping on tuples (you could ignore the second value) but we don't..
Let's do something else in the meantime and discuss tomorrow

rvanasa:
[...]

PR #168, opened 2025-02-21
Implement missing imperative `Set` functions

rvanasa:
Adds the following functions to the `Set` module:

* `toPure()`
* `fromPure()`
* `addAll()`
* `deleteAll()` (see #160 for naming discussion)
* `retainAll()`


src/Set.mo
```diff
@@ -758,6 +757,102 @@ module {
...
+  /// Space: `O(1)` retained memory plus garbage, see the note below.
+  /// where `m` and `n` denote the number of elements in `set` and `iter`, respectively,
+  /// and assuming that the `compare` function implements an `O(1)` comparison.
+  public func deleteAll<T>(set : Set<T>, compare : (T, T) -> Order.Order, iter : Types.Iter<T>) {
+    for (element in iter) {
+      delete(set, compare, element)
+    }
+  };
+
+  /// Removes all values in `set` aside from those contained in `iter`.
```
crusso, line 817:
```suggestion
  /// Retains all values in `set` aside from those contained in `iter`.
```

src/Set.mo
```diff
@@ -758,6 +757,102 @@ module {
...
+  /// where `m` and `n` denote the number of elements in `set` and `iter`, respectively,
+  /// and assuming that the `compare` function implements an `O(1)` comparison.
+  public func addAll<T>(set : Set<T>, compare : (T, T) -> Order.Order, iter : Types.Iter<T>) {
+    for (element in iter) {
+      add(set, compare, element)
+    }
+  };
+
+  /// Deletes all values in `iter` from the specified `set`.
+  /// This is different from `Set.intersect()` in that the function will trap if
```
crusso, line 789:
```suggestion
  /// This is different from `Set.diff()` in that the function will trap if
```
I think you mean Set.diff?

src/Set.mo
```diff
@@ -758,6 +757,102 @@ module {
...
+  /// where `m` and `n` denote the number of elements in `set` and `iter`, respectively,
+  /// and assuming that the `compare` function implements an `O(1)` comparison.
+  public func deleteAll<T>(set : Set<T>, compare : (T, T) -> Order.Order, iter : Types.Iter<T>) {
+    for (element in iter) {
+      delete(set, compare, element)
+    }
+  };
+
+  /// Removes all values in `set` aside from those contained in `iter`.
+  /// This is equivalent to `Set.diff()` but modifies the set in place.
```
crusso, line 818:
```suggestion
  /// This is equivalent to `Set.intersect()` but modifies the set in place.
```
I think you mean intersect?

crusso (commented):
Curious, why do these take iterators not sets? For example, I think retainAll might be simpler and more efficient if you were given a set.

Don't go and rewrite. I'm just asking.

rvanasa:
> Curious, why do these take iterators not sets? For example, I think retainAll might be simpler and more efficient if you were given a set.

The idea is for these functions to have a consistent API if we eventually include them for other data structures. Perhaps we could have `unionInPlace()`, `intersectInPlace()`, etc. (or whatever we'd like to name them) for better performance in the case of having two `Set` values.

src/Set.mo
```diff
@@ -758,6 +757,102 @@ module {
...
+  ///   Debug.print(debug_show(Iter.toArray(Set.values(difference)))); // => [1, 2]
+  /// }
+  /// ```
+  ///
+  /// Runtime: `O(m * log(n))`.
+  /// Space: `O(1)` retained memory plus garbage, see the note below.
+  /// where `m` and `n` denote the number of values in `set` and `iter`, respectively,
+  /// and assuming that the `compare` function implements an `O(1)` comparison.
+  public func retainAll<T>(set : Set<T>, compare : (T, T) -> Order.Order, iter : Types.Iter<T>) {
+    // TODO: optimize?
```
crusso, line 840: Yeah, this looks quadratic to me.

a simple improvement might be to construct a local Set from iter so you can use contains not linear search for testing common membership.
Also  setList could just be created as an ordinary array since we know the size of set up-front. Probably cheaper.

rvanasa: How about we just remove `retainAll()` and then include `*InPlace()` functions which take two Set values?

rvanasa: I noticed that Rust, Kotlin, and a few other languages use a predicate instead of collection as an argument to this function, which seems a lot better in general. I'll update the PR to go with this, and we can revise from there.

rvanasa: Updated, although it's a bit wonky having to pass both the `compare` and `predicate` functions. Perhaps this is another reason for prioritizing dataclasses or another language feature to remove those as soon as possible.

src/Set.mo
```diff
@@ -46,59 +48,59 @@ module {
...
+  /// import PureSet "mo:base/pure/Set";
+  /// import Nat "mo:base/Nat";
+  ///
+  /// persistent actor {
+  ///   let set = Set.empty<Nat>();
+  ///   Set.add(set, Nat.compare, 1);
+  ///   Set.add(set, Nat.compare, 2);
+  ///   Set.add(set, Nat.compare, 3);
+  ///   let pureSet = Set.toPure(set);
+  ///   assert(PureSet.contains(pureSet, 1));
```
crusso, line 65:
```suggestion
  ///   assert(PureSet.contains(pureSet, Nat.compare, 1));
```

src/Set.mo
```diff
@@ -46,59 +48,59 @@ module {
...
+  /// import Set "mo:base/Set";
+  /// import Nat "mo:base/Nat";
+  ///
+  /// persistent actor {
+  ///   var pureSet = PureSet.empty<Nat>();
+  ///   pureSet := PureSet.add(pureSet, Nat.compare, 1);
+  ///   pureSet := PureSet.add(pureSet, Nat.compare, 2);
+  ///   pureSet := PureSet.add(pureSet, Nat.compare, 3);
+  ///   let mutableSet = Set.fromPure(pureSet);
+  //    assert(Set.contains(mutableSet, 1));
```
crusso, line 93:
```suggestion
  //    assert(Set.contains(mutableSet, Nat.compare, 1));
```

src/pure/Set.mo
```diff
@@ -69,11 +69,11 @@ module {
     todo()
   };
 
-  public func intersect<T>(set1 : Set<T>, set2 : Set<T>) : Set<T> {
+  public func intersection<T>(set1 : Set<T>, set2 : Set<T>) : Set<T> {
```
crusso, line 72: I'll rename in my branch too.

src/pure/Set.mo
```diff
@@ -69,11 +69,11 @@ module {
...
     todo()
   };
 
-  public func diff<T>(set1 : Set<T>, set2 : Set<T>) : Set<T> {
+  public func difference<T>(set1 : Set<T>, set2 : Set<T>) : Set<T> {
```
crusso, line 76: ditto

src/pure/Set.mo
crusso, line 72:
```suggestion
  public func intersection<T>(set1 : Set<T>, set2 : Set<T>, compare: (T, T) -> Types.Order) : Set<T> {
```

crusso (approved):

PR #169, opened 2025-02-21
Implement misc. functions

### LINKED ISSUE #93: Implementation: misc. functions
Size estimate: M

Add missing implementations and doc comments for remaining `todo()` functions.

rvanasa:
This PR implements and adds doc comments for all functions with a `todo()` placeholder aside from the yet-to-be-implemented data structures.

Resolves #93.


src/pure/List.mo
```diff
@@ -61,8 +60,27 @@ module {
...
+  /// import Nat "mo:base/Nat";
+  /// List.contains<Nat>(?(1, ?(2, ?(3, null))), Nat.equal, 2) // => true
+  /// ```
+  ///
+  /// Runtime: O(size)
+  ///
+  /// Space: O(1)
+  ///
+  /// *Runtime and space assumes that `equal` runs in O(1) time and space.
+  public func contains<T>(list : List<T>, equal : (T, T) -> Bool, item : T) : Bool {
```
ggreif, line 76: Interestingly `pure/List.mo` has no `equals()` while `pure/Queue.mo` does have one. Does the API validator miss out on this?

ggreif: Also, the signature of `pure/Queue.contains` needs a fixup.

ggreif (approved):
LGTM, with comments addressed.

src/pure/List.mo
rvanasa: At one point I think there was a discussion about removing the `equal()` function for some modules, but I think it makes sense to include this. Updated the PR to require `equal()` for data structure modules as well as adjusting `pure/Queue`. Feel free to change the API as needed.

ggreif (approved):
Reapproving

PR #170, opened 2025-02-24
pure `List`  test brought over from old `base`

ggreif:
Adopted modifications and ready for iteration.


src/pure/List.mo
```diff
@@ -15,7 +15,7 @@ import Order "../Order";
 import Result "../Result";
 import Types "../Types";
 
-module {
+module List /*FIXME: remove?*/ {
```
ggreif, line 18:
```suggestion
module {
```

src/pure/List.mo
```diff
@@ -720,7 +720,7 @@ module {
 
   /// Split the given list into chunks of length `n`.
   /// The last chunk will be shorter if the length of the given list
-  /// does not divide by `n` evenly.
+  /// does not divide by `n` evenly. Passing `n = 0` is not allowed.
```
crusso, line 723:
```suggestion
  /// does not divide by `n` evenly. Passing `n = 0` is not allowed.
```
Why not say `Traps if n = 0`.

src/pure/List.mo
```diff
@@ -740,7 +740,7 @@ module {
   /// Space: O(size)
   public func chunks<T>(list : List<T>, n : Nat) : List<List<T>> =
     switch (split(list, n)) {
-      case (null, _) null;
+      case (null, _) { debug assert n > 0; null };
```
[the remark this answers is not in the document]
ggreif: I had it that way, but maybe we should do it unconditionally.

ggreif, line 743:
```suggestion
      case (null, _) { assert n > 0; null };
```

crusso (approved):

PR #171, opened 2025-02-24
Queue implementation 

luc-blaeser:
Implementation of `Queue`, including:
* Unit tests
* Large-scaled randomized tests
* API documentation

* Implement conversion to and from `immutable/Queue`.


rvanasa (approved):
LGTM; thank you! 

Added an inconsequential suggestion which we could do later if more convenient.

PR #172, opened 2025-02-24
Stack Implementation

luc-blaeser:
Implementation of `Stack`, including:
* Unit tests
* API documentation

* Implement conversion to and from `pure/List` (NB. functional list uses stack order).


rvanasa (approved):
LGTM; thank you!

src/Stack.mo
```diff
@@ -69,121 +366,454 @@ module {
...
+  ///   // prints `[3, 2, 1]`
+  /// }
+  /// ```
+  ///
+  /// Runtime: O(n)
+  /// Space: O(n)
+  /// where `n` denotes the number of elements stored on the stack and
+  /// assuming that `format` has O(1) costs.
+  public func toText<T>(stack : Stack<T>, format : T -> Text) : Text {
+    var text = "[";
```
rvanasa, line 774: For consistency with @ggreif's pure Queue implementation:

```suggestion
    var text = "Stack[";
```

crusso (commented):
This stack implementation is essentially just a wrapper around functional lists (encoded using records, not pairs).

Since pairs are generally cheaper, why not just the same representation as List.Pure.List<T>?






[11:39](https://dfinity.slack.com/archives/D041GF3UZ61/p1740656359516969)
If so, it might be cheaper to just an option list (Pure.List<T> = ?(T,List<T>) (less allocation and time spent looking up fields)

crusso (changes requested):
Or at least some justification for choosing records over tuples.

luc-blaeser:
> Or at least some justification for choosing records over tuples.
> 
> This would also make the toPure fromPure conversions 0(1), FWIW.
> 
> I'm happy to make the change

Ah, I see, I chose record for better code readability, as I usually consider tuples less ideal (position has an implicit semantics). I am happy with either approach. Just no time to adjust PR.

crusso:
I will quickly do the experiment of using Pure List for the node type as a PR of this one.
Converting to draft to avoid merge. @rvanasa @luc-blaeser

rvanasa (approved):

PR #174, opened 2025-02-25
Implement new `Random` API

rvanasa:
**Progress:**

* [x] Add the [SFC64](https://numpy.org/doc/2.2/reference/random/bit_generators/sfc64.html) pseudorandom number generator implementation from the [`prng`](https://mops.one/prng) Mops package (permission pending)
* [x] Adapt https://github.com/dfinity/motoko-training/pull/439 to implement a uniform random range, accounting for the bias caused by using a modulo operation
* [x] Add tests
* [ ] Thoroughly check for statistical biases
* [ ] Mirror implementations from `Random` to `AsyncRandom`
* [ ] Add code examples in doc comments

Note that `intRange()` and `natRange()` trap if the difference between the start and end values exceeds `Nat64.maxValue`. We could eventually use a different algorithm to cover larger input numbers. 

Here is a [benchmark](https://github.com/andanteyk/prng-seiran?tab=readme-ov-file#comparison) which motivates the choice of SFC64 as the default pseudorandom algorithm.


src/Random.mo
```diff
@@ -8,63 +18,206 @@ module {
...
+  /// but should not be used for cryptographic purposes.
+  /// The seed blob's first 8 bytes are used to initialize the PRNG.
+  public func fast(seed : Nat64) : Random {
+    let prng = PRNG.sfc64a();
+    prng.init(seed);
+    Random(
+      func() {
+        // Generate 8 bytes directly from a single 64-bit number
+        let n = prng.next();
+        let bytes = VarArray.repeat<Nat8>(0, 8);
```
crusso, line 32:
```suggestion
        // TODO: optimize using Array.tabulate or even better: a new primitive
        let bytes = VarArray.repeat<Nat8>(0, 8);
```

src/Random.mo
```diff
@@ -8,63 +18,206 @@ module {
...
+    };
 
     /// Random choice between `true` and `false`.
     public func bool() : Bool {
-      todo()
+      switch (bitIter.next()) {
+        case (?bit) { bit };
+        case null {
+          iter := generator().vals();
+          switch (bitIter.next()) {
```
crusso, line 81:
```suggestion
          switch (bitIter.next()) {
```
Lucky we never specified that an iterator cannot produce ?v after already having produced null. I guess the alternative would be to assign a new bitIter, but I think this fine...

src/Random.mo
```diff
@@ -8,63 +18,206 @@ module {
...
+          }
+        }
+      }
     };
 
     /// Random `Nat8` value in the range [0, 256).
-    public func byte() : Nat8 {
-      todo()
+    public func nat8() : Nat8 {
+      switch (iter.next()) {
```
crusso, line 93: I guess this mean we will discard the remaining bits in bitIter, right?
Maybe just add a TODO to use them in future? Or extend the bitIter api to give you the remaining bits?

src/Random.mo
```diff
@@ -8,63 +18,206 @@ module {
...
 
-    /// Random `Float` value in the range [0, 1).
-    /// The resolution is maximal (i.e. all relevant mantissa bits are randomized).
-    public func float() : Float {
-      todo()
+    // Helper function which returns a uniformly sampled `Nat64` in the range `[0, toExclusive)`.
+    // Uses rejection sampling to ensure uniform distribution even when the range
+    // doesn't divide evenly into 2^64. This avoids modulo bias that would occur
+    // from simply taking the modulo of a random 64-bit number.
+    func uniform64(toExclusive : Nat64) : Nat64 {
```
crusso, line 111:
```suggestion
    func uniform64(toExclusive : Nat64) : Nat64 {
```
Doesn't this mean you can't generate 2^64-1?
[0,0) is empty so the function should probably trap when toExclusive = 1.

Maybe this is case where we want an toInclusive, instead. Then no need to trap, and all values exressible.

src/Random.mo
```diff
@@ -8,63 +18,206 @@ module {
...
+      if (toExclusive <= 1) {
+        return 0
+      };
+      // 2^64 - (2^64 % toExclusive) = (2^64-1) - (2^64-1 % toExclusive):
+      let cutOff = Nat64.maxValue - (Nat64.maxValue % toExclusive);
+      // 2^64 / toExclusive, with toExclusive > 1:
+      let multiple = Nat64.fromNat(Nat.pow(2, 64) / Nat64.toNat(toExclusive));
```
crusso, line 118: What this scaling by multiple doing? My arithemetic is shocking so it might be fine, but I'd kinda expect you do just to take the base 2 logarithm of toExclusive, divde by 8, and use that to determine how many bytes you need to consume to construct the candidate sample (before possible rejection).

src/Random.mo
```diff
@@ -8,63 +18,206 @@ module {
...
+      loop {
+        // Build up a random Nat64 from bytes
+        var number : Nat64 = 0;
+        for (_ in Nat8.range(0, 8)) {
```
crusso, line 122: A while loop might be faster (or iterating over the array [0,..,7])

src/Random.mo
```diff
@@ -8,63 +18,206 @@ module {
...
-      todo()
+    public func nat64Range(from : Nat64, toExclusive : Nat64) : Nat64 {
+      if (from > toExclusive) {
+        Runtime.trap("Random.nat64Range(): from > toExclusive")
+      };
+      uniform64(toExclusive - from) + from
+    };
+
+    public func intRange(from : Int, toExclusive : Int) : Int {
+      switch (Nat.fromInt(toExclusive - from)) {
```
crusso, line 146: Does Nat.fromInt return an option? That would be different from all the other conversions that trap.... I wonder if we should revisit that choice....

crusso (commented):
This looks very nice to me. Minor comments - most could just be todos for future optimization.

src/internal/PRNG.mo
```diff
@@ -0,0 +1,75 @@
...
+    ///
+    /// Example:
+    /// ```motoko
+    public func init(seed : Nat64) = init3(seed, seed, seed);
+
+    /// Initializes the PRNG state with a hardcoded seed.
+    /// No argument is required.
+    ///
+    /// Example:
+    public func initPre() = init(0xcafef00dbeef5eed);
```
crusso, line 45:
```suggestion
    public func initPre() = init(0xcafef00dbeef5eed);
```
You could probably re-use `cafebabe` in the Trump era. (Just a grumpy joke)

src/Random.mo
rvanasa: That's a good point. We would definitely want to change this if it was exposed in the public API. This was directly taken from the original Random implementation ([here](https://github.com/dfinity/motoko-base/blob/9709055b7d81b51e35ddfbba0fae9efe5d52188a/src/Random.mo#L78)).

src/Random.mo
rvanasa: @ggreif, do you know if this was intentional? I was thinking about this as well (resetting the bit mask/byte on refill), but perhaps there was a performance reason for this.

src/Random.mo
rvanasa: Sure; this makes sense given that it's a private function. It probably makes sense to keep `natRange()` exclusive while perhaps adding a `natRangeInclusive()` for consistency with the `*.range()` / `*.rangeInclusive()` naming convention.

src/Random.mo
rvanasa: This comes directly from [this PR](https://github.com/dfinity/motoko-training/pull/439/files). I didn't want to make any changes in case I accidentally introduced a bug, but perhaps this could work.

crusso: Oh, don't trust me. My arithmetic is crap.

rvanasa:
@crusso, thank you for the high-quality review! Updated the PR, and maybe we can merge once @ggreif gets a chance to take a look as well.

src/Random.mo
```diff
@@ -8,64 +18,207 @@ module {
...
+  /// but should not be used for cryptographic purposes.
+  /// The seed blob's first 8 bytes are used to initialize the PRNG.
+  public func fast(seed : Nat64) : Random {
+    let prng = PRNG.sfc64a();
+    prng.init(seed);
+    Random(
+      func() {
+        // Generate 8 bytes directly from a single 64-bit number
+        let n = prng.next();
+        // TODO: optimize using Array.tabulate or even better: a new primitive
```
ggreif, line 32: I totally agree that this should be a prim. `toNat` followed by `fromNat` is wasteful.

src/Random.mo
```diff
@@ -8,64 +18,207 @@ module {
...
 
   public class Random(generator : () -> Blob) {
+    var iter : Iter.Iter<Nat8> = Iter.empty();
+    let bitIter : Iter.Iter<Bool> = object {
+      var mask = 0x00 : Nat8;
+      var byte = 0x00 : Nat8;
+      public func next() : ?Bool {
+        if (0 : Nat8 == mask) {
+          switch (iter.next()) {
+            case null { null };
```
ggreif, line 61: No need to wrap a block around this...
```suggestion
            case null null;
```
Many other cases below.

src/Random.mo
```diff
@@ -8,64 +18,207 @@ module {
...
-      todo()
+    // Helper function which returns a uniformly sampled `Nat64` in the range `[0, max]`.
+    // Uses rejection sampling to ensure uniform distribution even when the range
+    // doesn't divide evenly into 2^64. This avoids modulo bias that would occur
+    // from simply taking the modulo of a random 64-bit number.
+    func uniform64(max : Nat64) : Nat64 {
+      if (max == 0) {
+        return 0
+      };
+      // if (max == 1) {
```
alexandru-uta, line 116: hope you don't mind me asking: is this commented out code a leftover?

src/Random.mo
ggreif: I think it was intentional. I was contemplating having a single-bit mask being shifted in the iterator and skipping to the next byte when empty (maybe I have even implemented this somewhere?). OTOH, it would reduce the resupplying by a factor of 8...

validation/api/api.lock.json
```diff
@@ -900,8 +900,8 @@
     "exports": [
       "public class AsyncRandom(generator : () -> async* Blob)",
       "public let blob : shared () -> async Blob",
-      "public func new(seed : Blob) : Random",
-      "public func newAsync() : AsyncRandom",
+      "public func crypto() : AsyncRandom",
+      "public func fast(seed : Nat64) : Random",
```
crusso, line 904: Would better names be cryptographic/pseudo?

src/Random.mo
rvanasa: This is a possible optimization for more efficient usage of the random entropy. I'll remove this for now, and we can evaluate in a follow-up PR.

validation/api/api.lock.json
rvanasa: I considered `pseudo` in place of `fast` but thought that `fast` might be a more immediately clear name, i.e., instead of `pseudo` we would probably just want to call it `sfc64()`.

We could perhaps rename `crypto()` to `cryptographic()`. The reasoning for `crypto()` is that it implies that you want to use this RNG for cryptography, cryptocurrencies, etc. in the theme of naming based on the intended use case. I think `crypto` is an acceptable shorthand for `cryptographic` in this context but would be open to other opinions on that.

crusso (approved):
LGTM

rvanasa:
Merging for the initial release and making changes in follow-up PRs.

PR #176, opened 2025-02-25
chore: implement Set.add/remove/insert(All)/delete(All), adapt retainAll

crusso:
Implement add/remove/insert/delete/removeAll/insertAll for impure sets:

Set
-----------------------------------------------------------------------------------
|                               | Imp       |  Pure                     |  comment                        |
|---------------------|----------|---------------------| ---------------------------|
|  add(m, c, e)           |  ()          |   Set<K,V>            |                                         |
|  remove(m, c, e)     |  ()          |   Set<K,V>            |                                         |
|  insert(m, c, e)        |  Bool     |  ?Set<K,V>     ?     |                                          |
|  delete(m, c, e)       |  Bool     |  ?Set<K, V>    ?     |                                          |
|  insertAll(m, c, iter)    |  Bool  |         n/a               |  true iff size changed        |
|  deleteAll(m, c, iter)   | Bool   |        n/a                |  true iff size changed        |
|  retainAll(m, c, pred)  | Bool  |    n/a                    | true iff size changed |


src/Set.mo
```diff
@@ -790,13 +851,49 @@ module {
...
   /// and assuming that the `compare` function implements an `O(1)` comparison.
-  public func deleteAll<T>(set : Set<T>, compare : (T, T) -> Order.Order, iter : Types.Iter<T>) {
+  public func deleteAll<T>(set : Set<T>, compare : (T, T) -> Order.Order, iter : Types.Iter<T>) : Bool {
+    var deleted = false;
     for (element in iter) {
-      delete(set, compare, element)
-    }
+      deleted := delete(set, compare, element) or deleted // order matters!
+    };
+    return deleted
```
ggreif, line 859:
```suggestion
    deleted
```

src/Set.mo
```diff
@@ -790,13 +851,49 @@ module {
...
+  /// Runtime: `O(m * log(n))`.
+  /// Space: `O(1)` retained memory plus garbage, see the note below.
+  /// where `m` and `n` denote the number of elements in `set` and `iter`, respectively,
+  /// and assuming that the `compare` function implements an `O(1)` comparison.
+  public func insertAll<T>(set : Set<T>, compare : (T, T) -> Order.Order, iter : Types.Iter<T>) : Bool {
+    var inserted = false;
+    for (element in iter) {
+      inserted := insert(set, compare, element) or inserted // order matters!
+    };
+    return inserted
```
ggreif, line 891:
```suggestion
    inserted
```

src/Set.mo
```diff
@@ -366,10 +391,12 @@ module {
     switch (insertResult) {
       case (#inserted) {
         // if inserted an element that was not previously there, increment the tree size counter
-        set.size += 1
+        set.size += 1;
+	return true
```
ggreif, line 395: these `return`s are in tail position, redundant

src/Set.mo
```diff
@@ -454,9 +517,7 @@ module {
         deletedElement
       }
     };
-    if (not deleted) {
-      Runtime.trap("Element is not present")
-    }
+    return deleted
```
ggreif, line 520:
```suggestion
    deleted
```

ggreif (approved):
LGTM. Commented on redundant `return`, but not insisting. (Compiler tends to add an extra instruction without need.) Also all `assert (` could be written without the outer parens. Oh well.

src/Set.mo
crusso, line 395:
```suggestion
	true
```

src/Set.mo
```diff
@@ -382,13 +409,14 @@ module {
           children
         });
         // promotion always comes from inserting a new element, so increment the tree size counter
-        set.size += 1
+        set.size += 1;
+	return true
```
crusso, line 413:
```suggestion
	true
```

src/Set.mo
```diff
@@ -366,10 +391,12 @@ module {
     switch (insertResult) {
       case (#inserted) {
         // if inserted an element that was not previously there, increment the tree size counter
-        set.size += 1
+        set.size += 1;
+	true
       };
       case (#existent) {
         // keep size
+	return false
```
crusso, line 399:
```suggestion
	false
```

crusso:
> LGTM. Commented on redundant `return`, but not insisting. (Compiler tends to add an extra instruction without need.) Also all `assert (` could be written without the outer parens. Oh well. I guess the `add` vs. `insert` and `delete` vs. `remove` dictionary will be a bit confusing to users.

I fixed the returns but lots of the existing assert have redundant parens. Let's not sweat the small stuff.. 

I actually agree that the add/insert distinction may be confusing but the return type will distinguish them.
Luc wanted trapping ones so we settled on boolean returning for applications that care, and non-boolean returning for convenience.

Let's see what external feedback brings.

test/Set.test.mo
```diff
@@ -25,6 +25,41 @@ run(
...
+          Iter.toArray(Set.values(set))
+	},
+        M.equals(T.array<Nat>(T.natTestable, []))
+      ),
+      test(
+        "delete empty",
+	do { let set = Set.empty<Nat>();
+	     assert (not Set.delete(set, Nat.compare, 0));
+             Iter.toArray(Set.values(set))
+	},
```
crusso, line 60:
```suggestion
	do { 
	   let set = Set.empty<Nat>();
	   assert (not Set.delete(set, Nat.compare, 0));
           Iter.toArray(Set.values(set))
	},
```

rvanasa (approved):

PR #179, opened 2025-02-26
Rewrite `pure/Queue` tests to use `test` package

rvanasa:
Improves the readability of the `pure/Queue` tests and console output in addition to a few minor fixes.

### DIFF
--- a/src/pure/Queue.mo
+++ b/src/pure/Queue.mo
@@ -24,9 +24,8 @@ import Iter "../Iter";
 import List "List";
 import Order "../Order";
 import Types "../Types";
-import { todo } "../Debug";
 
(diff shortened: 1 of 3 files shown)


src/pure/Queue.mo
```diff
@@ -24,9 +24,8 @@ import Iter "../Iter";
 import List "List";
 import Order "../Order";
 import Types "../Types";
-import { todo } "../Debug";
 
-module Queue /* FIXME */ {
+module Queue {
```
ggreif, line 28: I only needed this name for local testing in the interpreter, we can remove it.

ggreif (approved):
THX! Looks good!

PR #180, opened 2025-02-26
Improve code formatting CI check

rvanasa:
This PR removes [`pretty-quick`](https://www.npmjs.com/package/pretty-quick) (used in the original base library) in favor of directly using the Prettier CLI. It seems as though in practice, `pretty-quick` allows non-formatted code to accumulate over time. `npm run format:check` and `npm run format` will now apply to all Motoko files rather than only modified files.


ggreif (approved):
rubber-stamping

PR #185, opened 2025-02-27
Fix compiler warnings

rvanasa:
Fixes various compiler warnings across the codebase.

### DIFF
--- a/src/pure/List.mo
+++ b/src/pure/List.mo
@@ -97,7 +97,7 @@ module {
   /// Space: O(1)
   public func get<T>(list : List<T>, n : Nat) : ?T = switch list {
     case null null;
-    case (?(h, t)) if (n == 0) ?h else get(t, n - 1)
+    case (?(h, t)) if (n == 0) ?h else get(t, n - 1 : Nat)
   };
 
   /// Add `item` to the head of `list`, and return the new list.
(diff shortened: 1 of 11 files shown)


src/pure/List.mo
```diff
@@ -97,7 +97,7 @@ module {
   /// Space: O(1)
   public func get<T>(list : List<T>, n : Nat) : ?T = switch list {
     case null null;
-    case (?(h, t)) if (n == 0) ?h else get(t, n - 1)
+    case (?(h, t)) if (n == 0) ?h else get(t, n - 1 : Nat)
```
crusso, line 100: (FYA Another way to avoid these warning would be to add the type parameter in recursive call: e.g. get<T>(..))

test/pure/Map.prop.test.mo
```diff
@@ -141,7 +139,7 @@ func run_all_props(range : (Nat, Nat), size : Nat, map_samples : Nat, query_samp
               "get(add(add(m, c, k, v1), c, k, v2), c, k) == ?v2",
               func(m, k) {
                 let (v1, v2) = ("V1", "V2");
-                Map.get(Map.add(Map.add(m, c, k, v1), c, k, v2), c, k) == v2
```
crusso, line 144: Good catch!

test/pure/Map.prop.test.mo
```diff
@@ -284,7 +282,7 @@ func run_all_props(range : (Nat, Nat), size : Nat, map_samples : Nat, query_samp
                 let k = Map.keys(m);
                 let v = Map.values(m);
                 for (e in Map.entries(m)) {
-                  if (e.0 != k.next() or e.1 != v.next()) return false
+                  if (?e.0 != k.next() or ?e.1 != v.next()) return false
```
crusso, line 285: Good catch!

crusso (approved):

PR #186, opened 2025-02-27
Re-enable `Text` module tests

rvanasa:
Renames `Text.test.skip.mo` back to `Text.test.mo`.

Depends on #172.


src/Queue.mo
```diff
@@ -70,11 +70,11 @@ module {
     let iter = values(queue);
     var current = pureQueue;
     loop {
-      switch(iter.next()) {
+      switch (iter.next()) {
```
rvanasa, line 73: (Unrelated formatting changes)

PR #188, opened 2025-02-27
chore: remove `Cycles.add`

ggreif:
and describe how to use a parenthetical


ggreif (commented):
@rvanasa Running `prettier` locally like
```
npx prettier --check '**/*.mo' --ignore-path .gitignore
```
doesn't give me formatting errors at all... What is different in CI?

src/Cycles.mo
```diff
@@ -6,6 +6,15 @@
...
 /// **NOTE:** Since cycles measure computational resources, the value of  `balance()` can change from one call to the next.
 ///
+/// To attach an amount of cycles to be transferred in a call,
+/// that is, evaluation of a shared function call or `async`
+/// expression, one prefixes the expression with a parenthetical
+/// of the form `(with cycles = <amount>) <call>`.
+/// **NOTE:** Traps if the cycles specified would exceed `2 ** 128` cycles.
+///
+/// Upon the call, but not before, the amount of cycles is deducted from `balance()`.
+/// If this total exceeds `balance()`, the caller traps, aborting the call.
```
crusso, line 16:
```suggestion
/// Cycles can be transferred from the current actor to another actor with the evaluation of certain forms of expression.
/// In particular, the expression must be a call to a shared function, a call to a local function with an `async` return type, or a simple `async` expression. 
/// To attach an amount of cycles to an expression, simply prefix the expression `<exp> with `(with cycles = <amount>)`, that is `(with cycles = <amount>) <exp>`. 
///
/// **NOTE:** Attaching cycles will trap if the amount specified exceeds `2 ** 128` cycles.
///
/// Upon the call, but not before, the amount of cycles is deducted from `balance()`.
/// If this total exceeds `balance()`, the caller traps, aborting the call without consuming the cycles.
/// Note that attaching cycles to a call to a local function call or `async` expression just transfer cycles from the current actor to itself.
```

rvanasa:
> doesn't give me formatting errors at all... What is different in CI?

Are you running `npm run format` or that command directly? It's possible that npm is picking up an outdated version of either `prettier` or `prettier-plugin-motoko` in your global cache or `node_modules` directory. 

Fixed by updating the branch from main. Let me know if it happens again and we could look further into it.

ggreif (commented):
Thanks for the wordsmithing @crusso!

rvanasa (approved):

rvanasa (approved):

PR #189, opened 2025-02-27
chore: make `pure/List` mostly tail-recursive

ggreif:
TODO:
- [ ] `size` could be more efficient with a loop
- [x] `repeat` ditto
- [x] `tabulate` is not tail-recursive (but constructor tailed)
- [x] `concat` is not tail-recursive (but constructor tailed)
- [ ] `merge` is not tail-recursive (but constructor tailed)
- [x] `fromIter` is not tail-recursive (but constructor tailed)
- [ ] `reverse` looks tail recursive (could be more efficient with a loop)
- [x] `mapResult` needs another look (somewhat convoluted in legacy `base`)

### DIFF
--- a/src/pure/List.mo
+++ b/src/pure/List.mo
@@ -178,10 +178,11 @@ module {
   /// Space: O(size)
   ///
   /// *Runtime and space assumes that `f` runs in O(1) time and space.
-  public func forEach<T>(list : List<T>, f : T -> ()) = switch list {
-    case null ();
-    case (?(h, t)) { f h; forEach(t, f) }
-  };
+  public func forEach<T>(list : List<T>, f : T -> ()) =
+    switch list {
+      case (?(h, t)) { f h; forEach(t, f) };
+      case null ()
+    };
 
   /// Call the given function `f` on each list element and collect the results
   /// in a new list.
@@ -196,10 +197,11 @@ module {
   ///
   /// Space: O(size)
   /// *Runtime and space assumes that `f` runs in O(1) time and space.
-  public func map<T1, T2>(list : List<T1>, f : T1 -> T2) : List<T2> = switch list {
-    case null null;
-    case (?(h, t)) ?(f h, map(t, f))
-  };
+  public func map<T1, T2>(list : List<T1>, f : T1 -> T2) : List<T2> =
+    switch list {
+      case null null;
+      case (?(h, t)) ?(f h, map(t, f))
+    };
 
   /// Create a new list with only those elements of the original list for which
   /// the given function (often called the _predicate_) returns true.
--- a/src/pure/List.mo
+++ b/src/pure/List.mo
@@ -255,6 +255,18 @@ module {
   /// Space: O(size)
   ///
   /// *Runtime and space assumes that `f` runs in O(1) time and space.
+
+  public func mapResult<T, R, E>(list : List<T>, f : T -> Result.Result<R, E>) : Result.Result<List<R>, E> = (
+    func rev(acc : List<R>, list : List<T>, f : T -> Result.Result<R, E>) : Result.Result<List<R>, E> = switch list {
+      case null #ok acc;
+      case (?(h, t)) switch (f h) {
+        case (#ok fh) rev(?(fh, acc), t, f);
+        case (#err e) #err e
+      }
+    }
+  )(null, list, f) |> Result.mapOk(_, func(l : List<R>) : List<R> = reverse l);
+
+  /*
   public func mapResult<T, R, E>(list : List<T>, f : T -> Result.Result<R, E>) : Result.Result<List<R>, E> = switch list {
     case null #ok null;
     case (?(h, t)) {
@@ -263,7 +275,7 @@ module {
         case ((#err e, _) or (_, #err e)) #err e
       }
     }
-  };
+  };*/
 
   /// Create two new lists from the results of a given function (`f`).
   /// The first list only includes the elements for which the given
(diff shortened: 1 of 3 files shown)


test/pure/List.test.mo
```diff
@@ -1,3 +1,5 @@
+// @testmode wasi
```
rvanasa, line 1: Is this just to speed up the tests? We might want to run this with the interpreter (default) to check for stack overflows.

ggreif: To me it appeared that the interpreter was more tolerant in terms stack. Maybe we should have a `// @testmode both` directive, if not already possible.

src/pure/List.mo
```diff
@@ -804,16 +813,13 @@ module {
 
   public func toText<T>(list : List<T>, f : T -> Text) : Text {
     var text = "[";
-    var first = false;
     forEach(
       list,
       func(item : T) {
-        if first {
+        if (text.size() > 1) {
```
crusso, line 819: Are you sure text.size() is constant time when using ropes (in compiled code)? I don't recall off-hand. If not, revert!

src/pure/List.mo
```diff
@@ -58,10 +58,12 @@ module {
...
   /// Space: O(1)
-  public func size<T>(list : List<T>) : Nat = switch list {
-    case null 0;
-    case (?(_, t)) 1 + size t
-  };
+  public func size<T>(list : List<T>) : Nat = (
+    func go(n : Nat, list : List<T>) : Nat = switch list {
+      case (?(_, t)) go(n + 1, t);
+      case null n
+    }
```
crusso, line 65: A while loop would be cheaper (no inital function call). See Queue.fromPure for similar.

src/pure/List.mo
```diff
@@ -58,10 +58,12 @@ module {
...
+  )(0, list);
```
crusso, line 66:
```suggestion
  public func size<T>(list : List<T>) : Nat = {
     var size = 0;
     var tmp = list;
     loop { 
        switch tmp {
          case (?(_, tl)) {
             size += 1; 
             tmp := tl;
          };
          case null  return size
        }
   }
```
Is probably cheaper (no call to a helper).

crusso:
There's more TODO (see #195)

src/pure/List.mo
ggreif: Good point! I'll revert!

src/pure/List.mo
ggreif: Yep, I was hoping that the compiler could create the same (or better) code by TCO. I.e. not sure if `var` allocates in this example (it shouldn't).

ggreif: I did the same transformation in `repeat`, though.

src/pure/List.mo
crusso: (It's no 0(1), I checked)

src/pure/List.mo
crusso: reverse needs it too. TCO won't inline the auxialliary function so ....

src/pure/List.mo
ggreif: Wait, do ropes carry a size around? That would help certainly :-)

src/pure/List.mo
ggreif: I actually test `reverse` for stack overflow and I haven't seen any 😖

ggreif: But for TCO one only have to turn the body of the helper into a loop, no need to inline it into the wrapper. Or am I mistaken?

ggreif: Of course the interpreter won't do TCO, and if we want constant stack there we'll have to manually TCO anyway...

src/pure/List.mo
ggreif: Nice:
```
// Layout of a concat node:
//
// ┌────────────┬─────────┬───────┬───────┐
// │ obj header │ n_bytes │ text1 │ text2 │
// └────────────┴─────────┴───────┴───────┘
```

src/pure/List.mo
```diff
@@ -585,8 +586,13 @@ module {
...
   public func tabulate<T>(n : Nat, f : Nat -> T) : List<T> {
-    func go(at : Nat, n : Nat) : List<T> = if (n == 0) null else ?(f at, go(at + 1, n - 1));
-    go(0, n)
+    var i = 0;
+    var l : List<T> = null;
+    while (i < n) {
+      l := ?(f i, l);
+      i += 1
+    };
+    reverse l
```
ggreif, line 595: This is a shame. A clever optimisation would spot the linear heap allocation discipline arising from the recursive algorithm (did somebody say abstract interpretation?) and pre-allocate the list nodes. Then fill them in like an array...

crusso: Yes, that's called tail-modulo-cons optimization but we don't have it.

src/pure/List.mo
crusso: > I actually test `reverse` for stack overflow and I haven't seen any 😖 With depth 100_000!

If you test in the intrepeter you won't see because the interpreter is in CPS... On wasm, you should see

ggreif: I test on Wasm, and don't see it :-)

crusso: Sorry, you reverse is tail recursive, but it allocates and calls a helper that is probably slower than just using a loop and a temp, without the additional function call.

crusso:
Is this still a draft?

src/pure/List.mo
ggreif: Reverting, those are bytes, not chars.

rvanasa (approved):

PR #190, opened 2025-02-27
optimization: Implement imperative stack using tuple, not record, based lists.

crusso:
(As discussed offline with Luc)*

Should be cheaper, allows constant time conversion to pure List and not really that much less readable.

In future, we could probably reimplement most operations using the underlying List functions.

*well, online, but different line.

If approved, I'll merge into #172 and then merge #172 into master.


luc-blaeser (approved):
Great! Thank you

PR #193, opened 2025-02-28
Temporarily replace `.values()` with `.vals()` for compatibility with previous Motoko versions (no API change)

rvanasa:
Because ICP Ninja currently uses a version of Motoko preceding the `.values()` alias for the `.vals()` field of `Array` and `Blob`, this PR temporarily updates all code usages to reduce friction for community developers trying out the base library on previous Motoko versions. 

This has no effect on the API (just an implementation detail).


luc-blaeser:
Thanks for preparing this. Although, I am not sure. I believe the proper thing would be that Ninja updates their dependencies. Real use of new-base would be by end of March, as now is only beta release.

crusso (approved):

rvanasa:
It could make sense to do this regardless of Ninja to reduce friction for developers trying the new base library in local setups with previous Motoko versions. There's no effect for the external-facing API, so we can easily change this back with a find/replace.

PR #214, opened 2025-03-04
Set toolchain version to 0.13.7

rvanasa:
Reconfigures the test runner to use `moc` from the latest dfx version. Since the priority is collecting user feedback, we want to make sure that developers can easily use the new base library without having to use a custom `moc` installation.


ggreif (approved):
But what is the problem?

ggreif:
`dfx` 0.25.1 will have the latest `moc`, I hope.

ggreif (approved):

PR #219, opened 2025-03-05
Rename `List.push()` to `pushFront()` and `List.pop()` to `popFront()`

rvanasa:
Renames `List.push()` and `List.pop()` for clarity and consistency with `Queue`.

Another reason to disambiguate is that JS uses `push()` to append an element to the end of the array, whereas in this case it's prepended to the beginning of the list.


ggreif (approved):
I agree, good move!

PR #221, opened 2025-03-05
Add CI checks for `mops.toml`, readme, and changelog

rvanasa:
This PR adds several new CI jobs which ensure that the versions listed in the readme and changelog are consistent with the `mops.toml` file. It also enforces updating the changelog when modifying any `src/**/*.mo` file and otherwise skips running unit tests.


ggreif (approved):
Rubber-stamping!

PR #222, opened 2025-03-06
Add missing tests for `Char`, `Func`, `Iter` and convert to `mo:test` format

rvanasa:
This PR converts `Char`, `Func`, and `Iter` to use the `mo:test` library as well as adding missing tests for these modules.

Also removes various compiler warnings in `Map`, `Principal`, and `pure/Set`.

I also fixed a bug in `Iter.step()` (caught by the new tests), which is included in the changelog as part of this PR.


ggreif (approved):
Wild! How did `Iter.step()` compile at all?

PR #226, opened 2025-03-12
Improve `Random` module to allow persistence in stable memory

rvanasa:
Simplifies the `Random` and `AsyncRandom` classes by deferring state management to `nextBit` and `nextByte` input functions.

This makes it possible to store the state of `Random` / `AsyncRandom` instances in stable memory, which was not possible before due to bit iteration using an internally-managed state. 

The `Random(generator)` initialization prior to this PR remains available as `Random.fromGenerator(generator)`.

The performance impact of this change should be negligible or at least worth the extra flexibility.

### DIFF
--- a/validation/api/api.lock.json
+++ b/validation/api/api.lock.json
@@ -898,11 +898,13 @@
   {
     "name": "Random",
     "exports": [
-      "public class AsyncRandom(generator : () -> async* Blob)",
+      "public class AsyncRandom(",
       "public let blob : shared () -> async Blob",
       "public func crypto() : AsyncRandom",
       "public func fast(seed : Nat64) : Random",
-      "public class Random(generator : () -> Blob)"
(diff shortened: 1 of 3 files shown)


validation/api/api.lock.json
```diff
@@ -898,11 +898,13 @@
   {
     "name": "Random",
     "exports": [
-      "public class AsyncRandom(generator : () -> async* Blob)",
+      "public class AsyncRandom(",
```
crusso, line 901: Weird truncation, probably coze args don't fit on a line...

rvanasa: Thanks for pointing that out; it's for the reason you mentioned as well as the regex-based logic not being able to handle the open/close curly braces + nested parentheses in the record types. I'll address this in a separate PR, since this overlaps with the logic needed for the code coverage checker.

crusso (approved):

PR #229, opened 2025-03-17
Real-time immutable queue with const time operations

### LINKED ISSUE #149: (pure) Queue is an amortized queue. Nipkow has non-amortized functional queue that may serve us better
Should we postpone implementing pure/Queue until we decide on a good representation?

Kamirus:
Adds `pure/RealTimeQueue` module which implements "[Real-Time Double-Ended Queue](https://drops.dagstuhl.de/storage/00lipics/lipics-vol268-itp2023/LIPIcs.ITP.2023.29/LIPIcs.ITP.2023.29.pdf)" ([formalization code archive](https://www.isa-afp.org/sessions/real_time_deque/#RealTimeDeque.html)) to achieve worst-case `O(1)` const-time&space push/pop/peek operations.
This implementation avoids performance cliffs that were leading to unexpected cycle/fuel exhaustion.
However, this implementation has worse amortized performance than `pure/Queue`. The differences between all three queue implementations are demonstrated in the `Queues.bench.mo` [benchmark](https://github.com/dfinity/new-motoko-base/pull/229#:~:text=Different%20queue%20implementations).

Resolves #149 

TODOs:
- [x] `pure/RealTimeQueue` should be an alternative to the `pure/Queue`, match the following aspects:
  - [x] examples in docs
  - [x] semantics of operations (order of visiting elements should match the `pure/Queue`) - exception: `filter` and `filterMap` preserve the front-to-back order, while `pure/Queue` does not
  - [x] both should provide the same operations
- [x] Make a simple benchmark that is fast and illustrates the differences in perf between both immutable and the mutable queues


src/pure/Queue.mo
```diff
@@ -88,19 +104,56 @@ module {
...
+  public func contains<T>(queue : Queue<T>, eq : (T, T) -> Bool, item : T) : Bool = switch queue {
+    case (#empty) false;
+    case (#one(x)) eq(x, item);
+    case (#two(x, y)) eq(x, item) or eq(y, item);
+    case (#three(x, y, z)) eq(x, item) or eq(y, item) or eq(z, item);
+    case (#idles(((l1, l2), _), ((r1, r2), _))) List.contains(l1, eq, item) or List.contains(l2, eq, item) or List.contains(r2, eq, item) or List.contains(r1, eq, item); // note that the order of the right stack is reversed, but for this operation it does not matter
+    case (#rebal((_, big, small))) {
+      let (extraB, _, (oldB1, oldB2), _) = BigState.current(big);
+      let (extraS, _, (oldS1, oldS2), _) = SmallState.current(small);
+      // note that the order is one of the stacks is reversed (depending on the `direction` field), but for this operation it does not matter
```
crusso, line 146:
```suggestion
      // note that the order of one of the stacks is reversed (depending on the `direction` field), but for this operation it does not matter
```

src/pure/Queue.mo
```diff
@@ -129,336 +189,805 @@ module {
...
+      case (null, ?(x, null)) #one(x);
+      case (?(x, null), null) #one(x);
+      case (null, ?(x, ?(y, null))) #two(y, x);
+      case (?(x, null), ?(y, null)) #two(y, x);
+      case (?(x, ?(y, null)), null) #two(y, x);
+      case (null, ?(x, ?(y, ?(z, null)))) #three(z, y, x);
+      case (?(x, ?(y, ?(z, null))), null) #three(z, y, x);
+      case (?(x, ?(y, null)), ?(z, null)) #three(z, y, x);
+      case (?(x, null), ?(y, ?(z, null))) #three(z, y, x);
+      case _ (trap "Illegal smallqueue invocation")
```
crusso, line 756: Please make this  (and any other trap) consistent with style of other traps in new-motoko-base (I think they add the name of the lib and operation that trapped)

validation/api/api.lock.json
```diff
@@ -1328,14 +1328,15 @@
...
       "public func empty<T>() : Queue<T>",
-      "public func equal<T>(queue1 : Queue<T>, queue2 : Queue<T>, equal : (T, T) -> Bool) : Bool",
-      "public func filter<T>(queue : Queue<T>, f : T -> Bool) : Queue<T>",
+      "public func equal<T>(queue1 : Queue<T>, queue2 : Queue<T>, equality : (T, T) -> Bool) : Bool",
+      "public func filter<T>(queue : Queue<T>, predicate : T -> Bool) : Queue<T>",
       "public func filterMap<T, U>(queue : Queue<T>, f : T -> ?U) : Queue<U>",
       "public func forEach<T>(queue : Queue<T>, f : T -> ())",
-      "public func fromIter<T>(iter : Iter.Iter<T>) : Queue<T>",
+      "public func fromIter<T>(iter : Iter<T>) : Queue<T>",
+      "public func idlesInvariant<T>(((l, nL), (r, nR)) : (Idle<T>, Idle<T>)) : Bool",
```
crusso, line 1339: Do we need this be public?

src/pure/Queue.mo
```diff
@@ -6,51 +6,64 @@
...
 /// Construction: Create a new queue with the `empty<T>()` function.
 ///
-/// Note on the costs of push and pop functions:
-/// * Runtime: `O(1)` amortized costs, `O(n)` worst case cost per single call.
-/// * Space: `O(1)` amortized costs, `O(n)` worst case cost per single call.
-///
-/// `n` denotes the number of elements stored in the queue.
+/// Note that some operations that traverse the elements of the queue (e.g. `forEach`, `values`) preserve the order of the elements,
+/// whereas others (e.g. `map`, `contains`, `any`) do NOT guarantee that the elements are visited in any order.
+/// The order is undefined to avoid allocations, making these operations more efficient.
```
crusso, line 21: I wonder if that's the right tradeoff to make - users might unwittingly rely on ordering. Would it be hard to preserve the order?

src/pure/Queue.mo
```diff
@@ -88,19 +104,56 @@ module {
...
-  public func size<T>(queue : Queue<T>) : Nat {
-    debug assert queue.1 == List.size(queue.0) + List.size(queue.2);
-    queue.1
+  public func size<T>(queue : Queue<T>) : Nat = switch queue {
+    case (#empty) 0;
+    case (#one(_)) 1;
+    case (#two(_, _)) 2;
+    case (#three(_, _, _)) 3;
+    case (#idles((l, nL), (r, nR))) {
+      debug assert Stacks.size(l) == nL and Stacks.size(r) == nR;
```
crusso, line 113: I have a horrible feeling that dfx always enables --debug so these assertion will always run in user code.
Might be worth checking (build a default dfx project check if debug expressions have an effect)

crusso: Or maybe Luc remembers what the status with dfx is.

crusso (commented):
This looks very nice, but also hard to review on a small screen. I think I need to wait until I'm on my big monitor again to review properly. Added some interim comments.

Let's hope the numbers are good too!

src/pure/Queue.mo
Kamirus: Oh no, [yes they are executed by default](https://internetcomputer.org/docs/motoko/main/reference/language-manual/#debug). People would need to pass `--release` flag. One would argue that every production program should be compiled with this flag. IMO it is good to have those invariants checked. Maybe I'll check which ones are actually expensive to compute...

src/pure/Queue.mo
Kamirus: It's not hard to code, but inefficient. Traversing in order requires `O(n)` allocations (because about a half of elements are in a list in reversed order). I think it would also be confusing for users that their `contains` check eats so much memory 😄

Kamirus: Since there are no breaking changes related to that I've opted for 'undefined' order for these pure-like operations.
So that we could always change that to 'front-to-back' order in future if needed.
Users can always use the `Iter` module directly to get the desired order.
But it would be nice to ask others in @dfinity/languages what they would expect 😄 
I'm not sure what is the right way to go. Maybe removing these functions is the way to go, no confusion, but no functionality, just use `Iter`

validation/api/api.lock.json
Kamirus: good catch, should be private

Kamirus:
> This looks very nice, but also hard to review on a small screen. I think I need to wait until I'm on my big monitor again to review properly. Added some interim comments.
> 
> Let's hope the numbers are good too!

So the initial numbers are [here](https://github.com/dfinity/new-motoko-base/pull/235#issuecomment-2748372541). Look for the phrase "Compare pure/Queue with base:Deque", there are 3 benchmarks

src/pure/Queue.mo
Kamirus: I'd prefer to leave the `debug assert` invariants for safety and a form of extra documentation.
`mops bench` uses now the Release flag by default. And users should be advised in the docs to use that flag in production.

src/pure/RealTimeQueue.mo
```diff
@@ -0,0 +1,1026 @@
+/// Double-ended immutable queue with guaranteed `O(1)` push/pop operations (caveat: high constant factor).
```
crusso, line 1: Very nice doc!

src/pure/RealTimeQueue.mo
```diff
@@ -0,0 +1,1026 @@
...
+///
+/// A Queue is internally implemented as a real-time double-ended queue based on the paper
+/// "Real-Time Double-Ended Queue Verified (Proof Pearl)". The implementation maintains
+/// worst-case constant time `O(1)` for push/pop operations through gradual rebalancing steps.
+///
+/// Construction: Create a new queue with the `empty<T>()` function.
+///
+/// Note that some operations that traverse the elements of the queue (e.g. `forEach`, `values`) preserve the order of the elements,
+/// whereas others (e.g. `map`, `contains`, `any`) do NOT guarantee that the elements are visited in any order.
+/// The order is undefined to avoid allocations, making these operations more efficient.
```
crusso, line 36: Maybe we should state this for Queue too.

src/pure/RealTimeQueue.mo
```diff
@@ -0,0 +1,1026 @@
...
+  /// ```
+  ///
+  /// Runtime: `O(1)`.
+  ///
+  /// Space: `O(1)`.
+  public func size<T>(queue : Queue<T>) : Nat = switch queue {
+    case (#empty) 0;
+    case (#one(_)) 1;
+    case (#two(_, _)) 2;
+    case (#three(_, _, _)) 3;
```
crusso, line 133:
```suggestion
    case (#one _) 1;
    case (#two _) 2;
    case (#three _) 3;
```
Sad, but we don't optimize that.

crusso: ( I think)

src/pure/RealTimeQueue.mo
```diff
@@ -0,0 +1,1026 @@
...
+  ///   let queue = Queue.singleton<Nat>(42);
+  ///   assert Queue.size(queue) == 1;
+  /// }
+  /// ```
+  ///
+  /// Runtime: `O(1)`.
+  ///
+  /// Space: `O(1)`.
+  public func size<T>(queue : Queue<T>) : Nat = switch queue {
+    case (#empty) 0;
```
crusso, line 130: All the #empty .. #three cases are presumably rare for real queues. I wonder if testing these cases last might actually increase perf (not for this PR but for future optimization).

src/pure/RealTimeQueue.mo
```diff
@@ -0,0 +1,1026 @@
...
+  /// The real-time queue data structure can be in one of the following states:
+  ///
+  /// - `#empty`: the queue is empty
+  /// - `#one`: the queue contains a single element
+  /// - `#two`: the queue contains two elements
+  /// - `#three`: the queue contains three elements
+  /// - `#idles`: the queue is in the idle state, where `l` and `r` are non-empty stacks of elements fulfilling the size invariant
+  /// - `#rebal`: the queue is in the rebalancing state
+  public type Queue<T> = {
+    #empty;
```
crusso, line 58: https://dl.acm.org/doi/pdf/10.1145/165180.165225 actually doesn't distinguish the #empty...#three cases and just uses a finite list. I wonder if that might improve perf a little, but I guess these are rare cases...
Perhaps Nipkow did this so simplify his proof.

src/pure/RealTimeQueue.mo
```diff
@@ -0,0 +1,1026 @@
...
+  /// Runtime: `O(1)`.
+  ///
+  /// Space: `O(1)`.
+  public func size<T>(queue : Queue<T>) : Nat = switch queue {
+    case (#empty) 0;
+    case (#one(_)) 1;
+    case (#two(_, _)) 2;
+    case (#three(_, _, _)) 3;
+    case (#idles((l, nL), (r, nR))) {
+      debug assert Stacks.size(l) == nL and Stacks.size(r) == nR;
```
crusso, line 135:
```suggestion
      debug assert Stacks.size(l) == nL and Stacks.size(r) == nR;
```
I wonder if we should disable this assert since peope are probably all building with debug on.

src/pure/RealTimeQueue.mo
```diff
@@ -0,0 +1,1026 @@
...
+/// - Deque (double-ended queue) by using any combination of push/pop operations on either end.
+///
+/// A Queue is internally implemented as a real-time double-ended queue based on the paper
+/// "Real-Time Double-Ended Queue Verified (Proof Pearl)". The implementation maintains
+/// worst-case constant time `O(1)` for push/pop operations through gradual rebalancing steps.
+///
+/// Construction: Create a new queue with the `empty<T>()` function.
+///
+/// Note that some operations that traverse the elements of the queue (e.g. `forEach`, `values`) preserve the order of the elements,
+/// whereas others (e.g. `map`, `contains`, `any`) do NOT guarantee that the elements are visited in any order.
```
crusso, line 35:
```suggestion
/// whereas others (e.g. `map`, `contains`, `any`) do NOT guarantee that the elements are visited in any order.
```
I think `any` does now, does it not?

src/pure/RealTimeQueue.mo
crusso: Maybe you could check it in the tests instead?

src/pure/RealTimeQueue.mo
```diff
@@ -0,0 +1,1026 @@
...
+        #idles((l, nL), (r, nR))
+      } else {
+        // initiate the rebalancing process
+        let targetSizeL = nL - nR - 1 : Nat;
+        let targetSizeR = 2 * nR + 1;
+        debug assert targetSizeL + targetSizeR == nL + nR;
+        let big = #big1(Current.new(l, targetSizeL), l, null, targetSizeL);
+        let small = #small1(Current.new(r, targetSizeR), r, null);
+        let states = (#right, big, small);
+        let states6 = States.step(States.step(States.step(States.step(States.step(States.step(states))))));
```
crusso, line 265: For future consideration: this might be a red-herring, but function calls are expensive on the IC. I wonder if using an iterative, tail-recursive version of `State.step(*n*, state) : state` could actually save some cost over apply State.step n times.
The body would be compiled to a loop and a jump which might just be cheaper, thought there will be some overhead due to checking the loop-counter.
Of course, inlining State.step would be even better but we don't do that.

src/pure/RealTimeQueue.mo
Kamirus: Good catch, I have to update that list now, and I'll put a similar note on the pure/Queue for now.

src/pure/RealTimeQueue.mo
Kamirus: I've followed [this implementation](https://www.isa-afp.org/sessions/real_time_deque/#RealTimeDeque.html). The paper was not accurate enough.

src/pure/RealTimeQueue.mo
Kamirus: Actually it made it worse. 

|                            |   pure/Queue | pure/RealTimeQueue | mutable Queue |
| :------------------------- | -----------: | -----------------: | ------------: |
| Initialize with 2 elements |   3_571 (0%) |         2_742 (0%) |    3_401 (0%) |
| Push 500 elements          | 103_492 (0%) | 1_113_817 (+5.12%) |  243_585 (0%) |
| Pop front 2 elements       |  98_792 (0%) |         5_243 (0%) |    4_326 (0%) |
| Pop 150 front&back         | 106_545 (0%) |   453_232 (+2.93%) |  140_211 (0%) |

Both as a tail-rec function step and as an explicit while loop.

src/pure/RealTimeQueue.mo
```diff
@@ -0,0 +1,1026 @@
...
+      let i1 = ((?(element, ?(a, null)), null), 2);
+      let i2 = ((?(c, ?(b, null)), null), 2);
+      #idles(i1, i2)
+    };
+    case (#idles(l0, (r, nR))) {
+      let (l, nL) = Idle.push(l0, element); // enque the element to the left end
+      // check if the size invariant still holds
+      if (3 * nR >= nL) {
+        debug assert 3 * nL >= nR;
+        #idles((l, nL), (r, nR))
```
crusso, line 256:
```suggestion
        #idles((l, nL), (r, nR))
```
You could avoid an allocation here of (r, nR). Shame the compiler doesn't do it.

src/pure/RealTimeQueue.mo
crusso: Concurs with my experiments.

src/pure/RealTimeQueue.mo
```diff
@@ -0,0 +1,1026 @@
...
+      case (?(h, _)) h;
+      case (null) Stacks.unsafeFirst(old)
+    };
+
+    public func size<T>((_, extraSize, _, targetSize) : Current<T>) : Nat = extraSize + targetSize
+  };
+
+  /// The bigger end of the queue during rebalancing. It is used to split the bigger end of the queue into the new big end and a portion to be added to the small end. Can be in one of the following states:
+  ///
+  /// - `#big1(cur, big, aux, n)`: Initial stage. Using the step function it takes `n`-elements from the `big` stack and puts them to `aux` in reversed order. `#big1(cur, x1 .. xn : bigTail, [], n) ->* #big1(cur, bigTail, xn .. x1, 0)`. The `bigTail` is later given to the `small` end.
```
crusso, line 833:
```suggestion
  /// - `#big1(cur, big, aux, n)`: Initial stage. Using the step function it takes `n`-elements from the `big` stack and puts them to `aux` in reversed order. `#big1(cur, x1 .. xn : bigTail, [], n) ->* #big1(cur, bigTail, xn .. x1, 0)`. The `bigTail` is later given to the `small` end.
```
```suggestion
  /// - `#big1(cur, big, aux, n)`: Initial state. Using the step function it takes `n`-elements from the `big` stack and puts them to `aux` in reversed order. `#big1(cur, x1 .. xn : bigTail, [], n) ->* #big1(cur, bigTail, xn .. x1, 0)`. The `bigTail` is later given to the `small` end.
```
typo? or maybe phase?

src/pure/RealTimeQueue.mo
```diff
@@ -0,0 +1,1026 @@
...
+  /// ```
+  ///
+  /// Runtime: `O(1)` worst-case!
+  ///
+  /// Space: `O(1)` worst-case!
+  public func popFront<T>(queue : Queue<T>) : ?(T, Queue<T>) = switch queue {
+    case (#empty) null;
+    case (#one(x)) ?(x, #empty);
+    case (#two(x, y)) ?(x, #one(y));
+    case (#three(x, y, z)) ?(x, #two(y, z));
```
crusso, line 341: FTR I found that re-ordering these tests last helps a little (also in reverse and pushFront). See #306.

src/pure/RealTimeQueue.mo
```diff
@@ -0,0 +1,1026 @@
...
+        let big = #big1(Current.new(l, targetSizeL), l, null, targetSizeL);
+        let small = #small1(Current.new(r, targetSizeR), r, null);
+        let states = (#right, big, small);
+        let states6 = States.step(States.step(States.step(States.step(States.step(States.step(states))))));
+        #rebal(states6)
+      }
+    };
+    // if the queue is in the middle of a rebalancing process: push the element and advance the rebalancing process by 4 steps
+    // move back into the idle state if the rebalancing is done
+    case (#rebal((dir, big0, small0))) switch dir {
```
crusso, line 271:
```suggestion
    case (#rebal (dir, big0, small0)) switch dir {
```

src/pure/RealTimeQueue.mo
```diff
@@ -0,0 +1,1026 @@
...
+  /// Space: `O(1)` worst-case!
+  public func popFront<T>(queue : Queue<T>) : ?(T, Queue<T>) = switch queue {
+    case (#empty) null;
+    case (#one(x)) ?(x, #empty);
+    case (#two(x, y)) ?(x, #one(y));
+    case (#three(x, y, z)) ?(x, #two(y, z));
+    case (#idles(l0, (r, nR))) {
+      let (x, (l, nL)) = Idle.pop(l0);
+      if (3 * nL >= nR) {
+        ?(x, #idles((l, nL), (r, nR)))
```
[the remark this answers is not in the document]
crusso: (see #307)

src/pure/RealTimeQueue.mo
crusso: (see #307)

crusso (approved):
Ok, I think I've done all I can without understanding the actual algorithm!
Very nice work!

crusso:
I'll leave it up to you whether you want to apply the microoptimization in #307 or not (about 9-10% on bench)

rvanasa (approved):
Thanks @crusso for the in-depth review! Thinking that we could merge this and make follow-up changes in separate PRs if needed.

PR #233, opened 2025-03-20
Fix `Array.enumerate()` / `VarArray.enumerate()`

rvanasa:
Fixes a bug in Array / VarArray `enumerate()` (introduced in the new base library) and includes tests.


crusso (approved):
Did this just trap at the end before?

rvanasa:
> Did this just trap at the end before?

Yep, so this fix won't break anyone's code.

PR #238, opened 2025-03-24
`Stack` docs

rvanasa:


Kamirus (approved):

src/Stack.mo
```diff
@@ -51,14 +51,15 @@ module {
...
   /// import PureList "mo:base/pure/List";
+  /// import Iter "mo:base/Iter";
   ///
   /// persistent actor {
   ///   let mutableStack = Stack.empty<Nat>();
-  ///   Stack.push(mutableStack, 0);
-  ///   Stack.push(mutableStack, 1);
+  ///   Stack.push(mutableStack, 3);
   ///   Stack.push(mutableStack, 2);
+  ///   Stack.push(mutableStack, 1);
```
Kamirus, line 60: Build stacks in reverse to emphasize the reversed order.
Elements are pushed in 'reversed order' `[3, 2, 1]` so that the output is `[1, 2, 3]`.
I hope to draw the users attention with the initial reversed order, so that the reversing-behaviour is acknowledged
-- this was confusing for me working on the examples, especially for functions like `fromIter` and `tabulate` that produce values reversed compared to PureList

rvanasa:
LGTM; thanks! That is a good idea to further emphasize the reverse order in the examples.

PR #241, opened 2025-03-24
`Float` docs

rvanasa:


src/Float.mo
```diff
@@ -12,11 +16,11 @@
 ///
 /// Examples of numerical errors:
 ///   ```motoko
-///   0.1 + 0.1 + 0.1 == 0.3 // => false
+///   assert 0.1 + 0.1 + 0.1 != 0.3;
 ///   ```
 ///
 ///   ```motoko
-///  1e16 + 1.0 != 1e16 // => false
+///   assert 1e16 + 1.0 == 1e16;
```
ggreif, line 23:
```suggestion
///   assert not (1e16 + 1.0 != 1e16);
```

src/Float.mo
```diff
@@ -32,7 +36,7 @@
 ///   let y = 0.3;
 ///
 ///   let epsilon = 1e-6; // This depends on the application case (needs a numerical error analysis).
-///   Float.equal(x, y, epsilon) // => true
+///   assert Float.equal(x, y, epsilon);
```
ggreif, line 39: What we really need is `(with epsilon = 1e-8) x == y`.

PR #248, opened 2025-03-26
`Tuples` docs

rvanasa:


src/Tuples.mo
```diff
@@ -1,25 +1,15 @@
 /// Contains modules for working with tuples of different sizes.
 ///
-/// Usage: either import the whole module and use the `TupleN` modules, or import the specific `TupleN` module directly.
+/// Usage example:
 ///
 /// ```motoko
-/// import Tuples "mo:new-base/Tuples"; // imports all TupleN modules
-/// import Nat "mo:new-base/Nat";
-/// import Text "mo:new-base/Text";
+/// import { Tuple2; Tuple3 } "mo:base/Tuples";
```
rvanasa, line 6: I keep going back and forth on whether we want to use `Tuple` or `Tuples` here. I suppose it depends on whether we want to eventually include conversions or any other functions in the parent module.

Kamirus: Would reverting to a single module with numbered functions be better?
`Tuple.compare2` instead of `Tuple.Tuple2.compare`?

rvanasa: `Tuple2.compare` is more natural for how we've restructured other APIs, e.g. splitting `Array` and `VarArray` in place of functions such as `tabulate` and `tabulateVar`. 

Here is another possibility for consistency with how we handle fixed-size Int / Nat values:

```motoko
import Tuple2 "mo:base/Tuple2";
import Tuple3 "mo:base/Tuple3";

// Precedent:
import Nat32 "mo:base/Nat32";
import Nat64 "mo:base/Nat64";
```

Kamirus: That might be the best solution, yes. But we should probably do it another PR, yes? Though this branch's name might suggest otherwise :D

rvanasa: Haha, yeah this was originally going to be the PR for renaming the tuple module. How about we merge this and then possibly rename in a follow-up PR...

PR #252, opened 2025-03-26
Make `pure/List` and `pure/Queue` stack-safe

Kamirus:
- Improves the stack safety and performance of various `pure/List` operations.
  - Changed operations are `map`, `filter`, `filterMap`, `partition`, `take`, `split`, `chunks`, `join`, `flatten`, `foldRight`, `merge`, and `zipWith`.
  - These functions now use tail-recursive helper functions for improved stack safety.
  - On top of that `flatten` is no longer quadratic.
- Makes `pure/Queue` stack safe by using the improved `List.split` instead of the `takeDrop` helper.
- Adds new benchmarks and missing tests.

### Benchmarking:

* Introduced a new benchmark suite `PureListStackSafety.bench.mo` to evaluate the stack safety of various `pure/List` functions.
* The benchmark can later be used to track the performance of these operations. FYI @crusso : If we implement 'Tail Recursion Modulo Cons' we can use the benchmark to see how much we would gain.

### Testing:

* Added new tests using the `mo:test` framework for the `join` and `flatten` functions in `List.test.mo`.

### DIFF
--- a/src/pure/List.mo
+++ b/src/pure/List.mo
@@ -436,10 +436,12 @@ module {
   /// Space: O(1) heap, O(size(list)) stack
   ///
   /// *Runtime and space assumes that `combine` runs in O(1) time and space.
-  public func foldRight<T, A>(list : List<T>, base : A, combine : (T, A) -> A) : A = switch list {
-    case null base;
-    case (?(h, t)) combine(h, foldRight(t, base, combine))
-  };
+  public func foldRight<T, A>(list : List<T>, base : A, combine : (T, A) -> A) : A = (
+    func go(list : List<T>, base : A, combine : (T, A) -> A) : A = switch list {
+      case null base;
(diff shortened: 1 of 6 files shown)


src/pure/List.mo
```diff
@@ -258,13 +264,13 @@ module {
...
-      case null #ok acc;
       case (?(h, t)) switch (f h) {
         case (#ok fh) rev(?(fh, acc), t, f);
         case (#err e) #err e
-      }
+      };
+      case null #ok(reverse acc)
     }
-  )(null, list, f) |> Result.mapOk(_, func(l : List<R>) : List<R> = reverse l);
+  )(null, list, f);
```
ggreif, line 273: good point!

ggreif (approved):
Nice! Thanks for doing this!
Of course the reversing will add heap/cycles overhead but the safety is probably worth it.

PR #253, opened 2025-03-26
`pure/Map` docs

rvanasa:


crusso:
@rvanasa PTAL - I already merged so I can get started on impure maps...

PR #254, opened 2025-03-26
`pure/Queue` docs

### LINKED ISSUE #198: doc: missing for pure/Queue

rvanasa:
Resolves #198.

@Kamirus, since we intend to use your new implementation, maybe we could splice these changes into https://github.com/dfinity/new-motoko-base/pull/229. Alternatively, we could merge #229 first and then handle the merge conflicts once they arise later on. Up to you which you'd prefer.

### DIFF
--- a/src/pure/Queue.mo
+++ b/src/pure/Queue.mo
@@ -77,7 +77,7 @@ module {
   /// import Queue "mo:base/pure/Queue";
   ///
   /// persistent actor {
-  ///   let queue = Queue.singleton<Nat>(25);
+  ///   let queue = Queue.singleton(25);
   ///   assert Queue.size(queue) == 1;
   /// }
   /// ```
@@ -94,7 +94,7 @@ module {
   /// import Queue "mo:base/pure/Queue";
   ///
   /// persistent actor {
-  ///   let queue = Queue.singleton<Nat>(42);
+  ///   let queue = Queue.singleton(42);
   ///   assert Queue.size(queue) == 1;
   /// }
   /// ```
@@ -117,7 +117,7 @@ module {
   /// import Queue "mo:base/pure/Queue";
   ///
   /// persistent actor {
-  ///   let queue = Queue.pushFront(Queue.pushFront(Queue.empty<Nat>(), 2), 1);
+  ///   let queue = Queue.pushFront(Queue.pushFront(Queue.empty(), 2), 1);
   ///   assert Queue.peekFront(queue) == ?1;
   /// }
   /// ```
@@ -138,7 +138,7 @@ module {
   /// import Queue "mo:base/pure/Queue";
   ///
   /// persistent actor {
-  ///   let queue = Queue.pushBack(Queue.pushBack(Queue.empty<Nat>(), 1), 2);
+  ///   let queue = Queue.pushBack(Queue.pushBack(Queue.empty(), 1), 2);
   ///   assert Queue.peekBack(queue) == ?2;
   /// }
   /// ```
@@ -185,7 +185,7 @@ module {
   /// import Queue "mo:base/pure/Queue";
   ///
   /// persistent actor {
-  ///   let queue = Queue.pushFront(Queue.pushFront(Queue.empty<Nat>(), 2), 1);
+  ///   let queue = Queue.pushFront(Queue.pushFront(Queue.empty(), 2), 1);
   ///   assert Queue.peekFront(queue) == ?1;
   ///   assert Queue.peekBack(queue) == ?2;
   ///   assert Queue.size(queue) == 2;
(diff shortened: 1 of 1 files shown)


Kamirus:
@rvanasa sure! IMO it is better to finish the doc PRs before continuing with the queue. Especially because from the [performance benchmarks comparing all 3 queue](https://github.com/dfinity/new-motoko-base/pull/229#issuecomment-2751575944) implementations we might want to have both 'pure/Queue' implementations...

src/pure/Queue.mo
```diff
@@ -208,18 +231,14 @@ module {
...
-  ///   case (?result) {
-  ///     let removedElement = result.0; // 1
-  ///     let reducedQueue = result.1; // queue with element [2].
-  ///   }
+  /// import Runtime "mo:base/Runtime";
+  ///
+  /// persistent actor {
+  ///   let initial = Queue.pushBack(Queue.pushBack(Queue.empty<Nat>(), 1), 2);
+  ///   // initial queue with elements [1, 2]
+  ///   let ?(frontElement, remainingQueue) = Queue.popFront(initial) else Runtime.trap "Empty queue impossible";
```
Kamirus, line 239: @rvanasa Do you know why this line causes '[Error: misplaced stability modifier: allowed on var or simple let declarations only](https://github.com/dfinity/new-motoko-base/actions/runs/14108942253/job/39522267362?pr=254#step:5:43)' ? Is `let ... else ...` not supported?

rvanasa: From conversations with Claudio and Luc, my understanding is that this is a limitation with the implicit `stable` keyword on the destructured `?(frontElement, remainingQueue)` let binding. 

We've decided to use a `transient` keyword for these situations (`transient let ?(frontElement, remainingQueue) = ...`) to provide examples for training LLMs how to navigate this situation. It might be possible to simply permit this syntax without showing an error, although I haven't looked into whether that would be complex to implement.

Kamirus: ahhh, because this is a top-level definition, not a local `let`, right. Now it makes sense 😄  thanks

src/pure/Queue.mo
```diff
@@ -99,7 +99,7 @@ module {
   /// }
   /// ```
   ///
-  /// Runtime: `O(1)`.
+  /// Runtime: `O(1)` in Release profile (compiled with `--release` flag), `O(n)` otherwise.
```
Kamirus, line 102: because of `debug assert`

src/pure/Queue.mo
```diff
@@ -285,8 +335,45 @@ module {
     check(list, List.size list, null)
   };
 
+  /// Convert a queue to an iterator of its elements in front-to-back order.
+  ///
+  /// Performance note: Creating the iterator needs `O(size)` runtime and space!
```
Kamirus, line 340: @rvanasa @crusso I've updated docs on the Runtime&Space bounds for the current pure/Queue operations.
**Also**: Just like in my [Real-Time Queue implementation](https://github.com/dfinity/new-motoko-base/pull/229#discussion_r2010088069) there are discrepancies in whether Queue operations preserve the order when traversing elements or not.

For example: `map`, `filter`, `filterMap`, `contains` **do not** traverse elements in front-to-back order;
Whereas, `any`, `all` and `forEach` **do** preserve the order, but pay `Space: O(size)` price for that.

**Implementation note:** all of these operations can either be implemented via `Iter`, e.g.
```motoko
Queue.values(q) |> Iter.map(_, f) |> Queue.fromIter(_)
```
Which preserves the order because of `values`. **Or by applying the operation directly** to lists in the queue state, which for operations like `any`, `all`, `contains` saves on allocations, but the order is undefined.

In a separate issue we should address that. For simplicity it is probably better to **always preserve the front-to-back order**.

src/pure/Queue.mo
```diff
@@ -308,16 +372,16 @@ module {
...
+  ///   let queue = Queue.fromIter([1, 2, 3].values());
+  ///   let allGreaterThanOne = Queue.all<Nat>(queue, func n = n > 1);
+  ///   assert not allGreaterThanOne; // false because 1 is not > 1
+  /// }
   /// ```
   ///
   /// Runtime: O(size)
   ///
-  /// Space: O(1)
+  /// Space: O(size) as the current implementation uses `values` to iterate over the queue.
```
Kamirus, line 384: O(size) but preserves order -- Do we even want more details like that ^ in the docs?

src/pure/Queue.mo
```diff
@@ -422,14 +509,30 @@ module {
...
+  /// Space: O(size)
   public func toText<T>(queue : Queue<T>, f : T -> Text) : Text {
     var text = "PureQueue[";
     func add(item : T) {
       if (text.size() > 10) text #= ", ";
       text #= f(item)
     };
     List.forEach(queue.0, add);
-    List.forEach(queue.2, add);
+    List.forEach(List.reverse(queue.2), add);
```
Kamirus, line 535: Bug, I'll open a separate PR with it

src/pure/Queue.mo
rvanasa: I'm in favor of always preserving the front-to-back order, since this is more likely to be intuitive for new developers.

src/pure/Queue.mo
rvanasa: It seems reasonable to add more details for any possibly unexpected time or space complexity.

rvanasa (commented):
Thanks for combing through the docs and finding that bug! Feel free to merge when ready.

PR #258, opened 2025-03-26
`Set` docs

rvanasa:
Starting point (LLM had difficulties updating this module).

### DIFF
--- a/src/Set.mo
+++ b/src/Set.mo
@@ -12,6 +12,7 @@
 ///   Set.add(userIds, Nat.compare, 1);
 ///   Set.add(userIds, Nat.compare, 2);
 ///   Set.add(userIds, Nat.compare, 3);
+///   assert Set.size(userIds) == 3;
 /// }
 /// ```
 ///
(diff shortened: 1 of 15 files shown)


src/Set.mo
```diff
@@ -1777,7 +1749,7 @@ module {
           case (#deleted) { #deleted };
           case (#inexistent) { #inexistent };
           case (#mergeChild({ internalChild })) {
-            #mergeChild({ internalChild; deletedElement = deleteElement })
+            #mergeChild({ internalChild })
```
rvanasa, line 1752: (Fixed a compiler warning)

crusso: Was this just a spurious, unexpected record field caught by Gabor's new warning?

crusso (approved):
LGTM (but I wrote it). @rvanasa PTAL and merge if happy

rvanasa (commented):
Thanks! Fixed two imports referencing `pure/Set` in the `Set` module. Otherwise looks good! Merging.

PR #262, opened 2025-03-28
Add `nullable()` and `fromNullable()` to `Array` and `VarArray`

rvanasa:
Provides a way to use (var-)arrays in a way that is familiar from Java / C#. This covers use cases which were originally handled by the `Buffer` class.


Kamirus:
I like the `fromNullable` idea, but I think the main use case is converting a `[var ?T]` into `[T]`, which is not covered in this PR. I'd like to take a step back and discuss use cases first. 

I see two use cases behind `Buffer` that are worth preserving:
1. "I want to gradually build an array and I know the exact size of it". What happens internally:
  - Initialize `[var ?T]` with `null`s and a given size
  - Set the items using `:=`
  - Finalize by getting `[T]` out of it
2. "I want to gradually build an array but I **don't** know the exact size of it. I'd like to choose the initial capacity (possibly 0) in order to avoid/reduce resizing costs. I'd like to add items either by using `add`-like method or by index."

API of the `Buffer` covers both use-cases. `List` offers better resizing, but cannot be initialized with a continuous array of given capacity.

**Ad 1.** It might be useful supporting the 1st use case with `VarArray` and functions introduced in this PR. But it would need another function : `[var ?T] -> ?[T]`, ideally in the `VarArray` module.

**Ad 2.** In order to support the 2nd use case as well, I'd explore a hybrid solution between `Buffer` and `List`. It would contain a fixed `[var ?T]` of given capacity and a resizable `List` tail for extra elements that exceed the initial capacity. 
Maybe in this way we could preserve the `Buffer` API while still improving the resizing performance.
I'd suggest naming it `ArrayBuilder` to indicate its purpose.

PR #270, opened 2025-04-06
Use `Order` type instead of the inlined variant

Kamirus:
Replaces every occurrence of `{ #less; #equal; #greater }` with the `Order.Order` type alias


ggreif (approved):
LGTM!

crusso (commented):
A more minimalist change might have to been to just refer to Types.Order. 
In the old base, I think we in-lined the types to avoid excessive library dependencies 
(and circular ones with Iter<T>)

PR #272, opened 2025-04-07
Add `Map.entriesFrom`

### LINKED ISSUE #271: Add a range function for `Map`
Why is it important? For example, in a DeFi application we could have a `K = (Principal, Transaction)` and we would want to obtain all `Transaction`'s that belong to a given `Principal`. But currently in the worst case we would need to parse all `entries` to find the relevant ones.

For example, it could look like this.

```motoko
public func entriesFrom<K, V>(
  map : Map<K, V>,
  compare : (K, K) -> Order.Order,
  key : K
) : Types.Iter<(K, V)>
```

* For a map `1 -> 'A', 2 -> 'B', 3 -> 'C'` and call `entriesFrom(map, compare, 2)` this function would return an iterator at `2 -> 'B'` .
* For a map `1 -> 'A', 3 -> 'C'` and the same call this function would return an iterator at `3 -> 'C'` 

This interface is quite similar to what is implemented in Rust [`ic_stable_structures`](https://github.com/dfinity/stable-structures/blob/2c30a85b467dcc2e4414de3216c4d053e434ee15/src/btreemap.rs#L1015). See also [Roman's blogpost](https://mmapped.blog/posts/14-stable-structures#stable-btree).

Kamirus:
Closes #271

### DIFF
--- a/src/Map.mo
+++ b/src/Map.mo
@@ -756,6 +756,19 @@ module {
     }
   };
 
+  public func entriesFrom<K, V>(
+    map : Map<K, V>,
+    compare : (K, K) -> Order.Order,
+    key : K
+  ) : Types.Iter<(K, V)> {
+    switch (map.root) {
+      case (#leaf(leafNode)) { return leafEntriesFrom(leafNode, compare, key) };
+      case (#internal(internalNode)) {
+        internalEntriesFrom(internalNode, compare, key)
+      }
+    }
+  };
+
   /// Returns an iterator over the key-value pairs in the map,
   /// traversing the entries in the descending order of the keys.
   ///
@@ -1369,6 +1382,24 @@ module {
     }
   };
 
+  func leafEntriesFrom<K, V>({ data } : Leaf<K, V>, compare : (K, K) -> Order.Order, key : K) : Types.Iter<(K, V)> {
+    var i = switch (BinarySearch.binarySearchNode<K, V>(data.kvs, compare, key, data.count)) {
+      case (#keyFound(i)) i;
+      case (#notFound(i)) i
+    };
+    object {
+      public func next() : ?(K, V) {
+        if (i >= data.count) {
+          return null
+        } else {
+          let res = data.kvs[i];
+          i += 1;
+          return res
+        }
+      }
+    }
+  };
+
   func reverseLeafEntries<K, V>({ data } : Leaf<K, V>) : Types.Iter<(K, V)> {
     var i : Nat = data.count;
     object {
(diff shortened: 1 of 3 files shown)


src/Map.mo
```diff
@@ -756,6 +756,42 @@ module {
...
+  /// ```motoko
+  /// import Map "mo:base/Map";
+  /// import Nat "mo:base/Nat";
+  /// import Iter "mo:base/Iter";
+  ///
+  /// persistent actor {
+  ///   let map = Map.empty<Nat, Text>();
+  ///   Map.add(map, Nat.compare, 0, "Zero");
+  ///   Map.add(map, Nat.compare, 1, "One");
+  ///   Map.add(map, Nat.compare, 3, "Three");
```
crusso, line 772:
```suggestion
  ///   let map = Map.fromIter([(0, "Zero"), (3, "Three),  (1, "One")].values(), Nat.compare)
```
(For consistency with my intended doc changes PR)

test/Map.test.mo
```diff
@@ -1543,4 +1545,84 @@ run(
...
+        check(3, [(2, "2"), (1, "1")]);
+        check(4, [(4, "4"), (2, "2"), (1, "1")]);
+        check(5, [(4, "4"), (2, "2"), (1, "1")])
+      }
+    );
+    Test.test(
+      "Extensive 2D test",
+      func() {
+        let map = Map.empty<Nat, Text>();
+        let n = 100;
```
crusso, line 1616: Maybe it would be worth testing manually with n = 1000 for more internal nodes... but perhaps that's too slow...

test/Set.test.mo
```diff
@@ -1907,4 +1909,83 @@ run(
...
+        check(3, [2, 1]);
+        check(4, [4, 2, 1]);
+        check(5, [4, 2, 1])
+      }
+    );
+    Test.test(
+      "Extensive 2D test",
+      func() {
+        let set = Set.empty<Nat>();
+        let n = 100;
```
crusso, line 1979: Same here, try with n = 1000

crusso (approved):
Nice,

Mostly cosmetic review since I'm not familiar with the code but you've got nice tests. 
Perhaps @luc-blaeser could take a deeper review.

I guess none of these new iterators are safe to concurrent modification, but I don't think the existing one are either.

crusso:
Do we want to add similar operations to pure/Set and pure/Map or wait until someone asks for those?

test/Set.test.mo
Kamirus: I did test manually with higher `n` in both cases. Also with smaller `btreeOrder` like `4` to cover more tree-shapes. Unfortunately the order is not customizable, otherwise I'd leave them in tests.
And yes I left `n=100` so that it's faster but still `>32`

Kamirus:
We could add them, but definitely in another PR

src/Set.mo
```diff
@@ -693,11 +689,8 @@ module {
   /// import Iter "mo:base/Iter";
   ///
   /// persistent actor {
-  ///   let set = Set.empty<Nat>();
-  ///   Set.add(set, Nat.compare, 0);
-  ///   Set.add(set, Nat.compare, 1);
-  ///   Set.add(set, Nat.compare, 3);
-  ///
+  ///   let set = Set.fromIter([0, 1, 3].values(), Nat.compare);
+  ///   assert Iter.toArray(Set.reverseValuesFrom(set, Nat.compare, 1)) == [3, 1];
```
christoph-dfinity, line 693: Looks to have duplicated the test below?
```suggestion
```

Kamirus: Yes thanks!

PR #278, opened 2025-04-09
Workflow: Make `validate-changelog` and skipping tests work again

Kamirus:
Workflow changes:
1. `validate-changelog` now correctly fails when there are no `Changelog.md` changes but there are `src/*.mo` changes -- no changes in logic, just fixing the existing one
2. Similar fix applied to skipping tests when there are no *.mo changes
3. Added skipping benchmarks as well

Details:
- Fixes how to get the correct `HEAD` commit reference in a workflow
- Corrects the skipping tests step to actually prevent the next step from executing
- Fetches the full git history where it is necessary

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -6,6 +6,9 @@ on:
   pull_request:
 
 jobs:
+  check-motoko-changes:
+    uses: ./.github/workflows/check-motoko-changes.yml
+
   validate-changelog:
     if: github.event_name == 'pull_request' && github.event.pull_request.base.ref == 'main'
     runs-on: ubuntu-24.04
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -62,11 +70,16 @@ jobs:
...
       - name: Check for Motoko file changes in 'src' directory
         id: check_mo_changes
         run: |
-          if ! git diff --name-only ${{ github.event.pull_request.base.sha || 'HEAD^' }} HEAD | grep -q '\.mo$'; then
+          BASE_SHA="${{ github.event.pull_request.base.sha || '$(git rev-parse HEAD~1)' }}"
+          HEAD_SHA="${{ github.event.pull_request.head.sha }}"
+
+          if ! git diff --name-only $BASE_SHA $HEAD_SHA | grep -q '\.mo$'; then
             echo "No Motoko file changes in src directory detected, skipping tests"
             exit 0
```
Kamirus, line 84: this is still doing nothing, the other steps still continue executing

rvanasa (approved):
Thank you for looking into this and finding a solution!

PR #282, opened 2025-04-10
Improve consistency of data structure `toText()` functions

### LINKED ISSUE #155: toText representation for Map
Currently implemented as
{}
{(k, v)}
{(k, v), ..., (k,v)}

to echo mathematical set notation and so we can easily copy (some) contents back into a Motoko expression (eg. Map.fromIter<K,V>(Map.fromArray [ ... ], K.compare))

Ryan suggests:

{}
{ k = v }
{ k = v; ... ; k = v }

(similar to JS, Python)

Note our keys may be complicated, not just atomic strings or numbers, so not sure the second notation scales that well.

I wonder if should (in future) provide a well-specified toJson/fromJson functionality...

### LINKED ISSUE #191: inconsistency: textual formats
Some libs prefix the output with data structure name "Queue[  ...]", others do not (maps and sets). I guess we should fix on a common pattern.

rvanasa:
Resolves #191.
Resolves #155 (indirectly).


Kamirus (approved):

crusso:
NB: we'll need to update the examples on the ai-doc branch...

PR #288, opened 2025-04-11
`VarArray` docs

rvanasa:
Merges into the `Array` docs PR so that it's quicker to compare the documentation between the two modules.


Kamirus (approved):
LGTM!

PR #294, opened 2025-04-14
chore: replace `Set.toText` by `Iter.toArray(Set.toIter ...)` in examples

crusso:
For consistency with Map docs.


ggreif (approved):
Looks canonical to me!

Kamirus (approved):
That is a good idea to avoid `Text` in asserts!

rvanasa (approved):
LGTM; thank you!

PR #296, opened 2025-04-15
`Int*` / `Nat*` docs

rvanasa:


alexandru-uta (approved):
looks good to me, thanks! (but there are so many one-liners that I can't guarantee I didn't miss anything)

PR #300, opened 2025-04-22
Trigger benchmarks when `mops.toml` changes

Kamirus:
* Modify `benchmarks.yml` to trigger benchmarks if either Motoko changes or mops dependency changes are detected.
* Update `check-motoko-changes.yml` to include detection of `mops.toml` changes. Refactor existing checks to avoid `grep` when looking for specific files in `git diff`.

### DIFF
--- a/.github/workflows/check-motoko-changes.yml
+++ b/.github/workflows/check-motoko-changes.yml
@@ -12,6 +12,9 @@ on:
       has_changelog_changes:
         description: "Whether Changelog.md was modified"
         value: ${{ jobs.check.outputs.has_changelog_changes }}
+      has_mops_dependency_changes:
+        description: "Whether mops dependency versions were changed"
+        value: ${{ jobs.check.outputs.has_mops_dependency_changes }}
 
 jobs:
   check:
@@ -20,6 +23,7 @@ jobs:
       has_mo_changes: ${{ steps.check_mo_changes.outputs.has_mo_changes }}
       has_src_mo_changes: ${{ steps.check_mo_changes.outputs.has_src_mo_changes }}
       has_changelog_changes: ${{ steps.check_mo_changes.outputs.has_changelog_changes }}
+      has_mops_dependency_changes: ${{ steps.check_mo_changes.outputs.has_mops_dependency_changes }}
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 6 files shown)


.github/workflows/check-motoko-changes.yml
```diff
@@ -54,3 +58,11 @@ jobs:
           else
             echo "has_changelog_changes=true" >> $GITHUB_OUTPUT
           fi
+
+          # Check for mops dependency changes
+          if ! git diff --name-only $BASE_SHA $HEAD_SHA | grep -q '^mops\.toml$'; then
```
ggreif, line 63: Why not tell `git` to focus on that file?
```suggestion
          if ! git diff --name-only $BASE_SHA $HEAD_SHA -- mops.toml
```
```suggestion
          if ! git diff --name-only $BASE_SHA $HEAD_SHA | grep -q '^mops\.toml$'; then
```

Kamirus: Good idea

.github/workflows/check-motoko-changes.yml
```diff
@@ -32,25 +36,33 @@ jobs:
...
+          if [ -z "$(git diff --name-only $BASE_SHA $HEAD_SHA -- '*.mo')" ]; then
             echo "has_mo_changes=false" >> $GITHUB_OUTPUT
             echo "No Motoko file changes detected"
           else
             echo "has_mo_changes=true" >> $GITHUB_OUTPUT
           fi
 
           # Check for src/*.mo changes
-          if ! git diff --name-only $BASE_SHA $HEAD_SHA | grep -q '^src/.*\.mo$'; then
+          if [ -z "$(git diff --name-only $BASE_SHA $HEAD_SHA -- 'src/*.mo')" ]; then
```
ggreif, line 47: Does really work? TIL, you can specify patterns there (or is that just shell expansion?). But then do we still need `--name-only`?

ggreif (approved):
Looks good to me!

.github/workflows/check-motoko-changes.yml
Kamirus: `--name-only` is not necessary, just makes for a shorter output.
`git diff` always returns 0, so I needed to compare its output whether it's empty.
Yes, patterns like `'src/*.mo'` work, this finds all `.mo` files under the `src/` directory in the root project

PR #301, opened 2025-04-22
Replace `Set.toText` by `Iter.toArray(Set.toIter ...)` in examples

rvanasa:
Extracts changes from https://github.com/dfinity/new-motoko-base/pull/294 which seem to be lost in the noise from merging #237 and #224.


kritzcreek (approved):

christoph-dfinity (approved):
LGTM

PR #303, opened 2025-04-29
CI: Enable docstring example validation

Kamirus:
This PR:
- Enables `npm run validate:docs` for changed `src` files.
- Fixes broken examples in various modules.

---

TODO / Future work:
- We definitely want to check the examples in the docstrings, however currently this step:
  - takes too long to always check every file, (why is it significantly slower than the unit tests? Maybe it could be easily parallelized?) For now `validate:docs` is run only for changed files
- Consider running the full validation on push to main

Broken docs:

- [x] src/List.mo ❌ 6
- [x] src/Map.mo ❌ 2
- [x] src/Set.mo ❌ 18
- [x] src/pure/List.mo ❌ 2
- [x] src/pure/Set.mo ❌ 1

1161 passed, 29 failed, 0 skipped


crusso (approved):
LGTM, but I'm not a GH actions expert. 
Looks like it's complaining about it's own Changelong entry...

.github/workflows/tests.yml
```diff
@@ -55,6 +56,19 @@ jobs:
...
+    if: needs.check-motoko-changes.outputs.has_src_mo_changes == 'true'
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup
+      - name: Validate docs for changed src/*.mo files
+        run: |
+          echo "${{ needs.check-motoko-changes.outputs.src_mo_changed_files }}" | while read -r file; do
+            npm run validate:docs "$file"
+          done
```
rvanasa, line 70:
```suggestion
        run: |
          echo "${{ needs.check-motoko-changes.outputs.src_mo_changed_files }}" | while read -r file; do
            npm run validate:docs "$file"
          done
        env:
          FORCE_COLOR: 2 # https://github.com/chalk/supports-color/issues/106
```

rvanasa (commented):
Thanks! We originally has this in CI, but it was apparently lost in the shuffle while merging into the main branch:

[Original](https://github.com/dfinity/new-motoko-base/commit/da66a3135e24c44f4043a9ffa9095ad504cb9870#diff-a5c0f08bc43f53a0451cad4a4a2e435ce55ad34183350dfae5e4b2c7dca6360b):

```yml
  docs:
    runs-on: ${{ matrix.os }}
    strategy:
      fail-fast: true
      matrix:
        os: [ubuntu-24.04]
        node: [20]
    steps:
      - uses: actions/checkout@v4
        with:
          fetch-depth: "0"
      - uses: actions/setup-node@v4
        with:
          node-version: ${{ matrix.node }}
      - run: npm ci
      - name: Validate changed files
        run: |
          CHANGED_FILES=$(git diff --name-only ${{ github.event.pull_request.base.sha || 'origin/main' }} HEAD | grep '^src/.*\.mo$' || true)
          if [ -n "$CHANGED_FILES" ]; then
            echo "$CHANGED_FILES" | xargs npm run validate:docs
          else
            echo "No Motoko files changed"
          fi
        env:
          FORCE_COLOR: 2 # https://github.com/chalk/supports-color/issues/106
```

Maybe we could run the full docs validation for release PRs, i.e. those which modify the version in the `package.json` file.

rvanasa (approved):

PR #305, opened 2025-05-02
CI: validate all docs on version change

Kamirus:
Adds full `validation:docs` on version change in `mops.toml`.
Continuation after #303 FYI @rvanasa .

Manual test: version change triggered the [workflow](https://github.com/dfinity/new-motoko-base/actions/runs/14790146639/job/41525569513?pr=305), it took almost 24mins!


rvanasa (approved):
LGTM; thanks! It will be great to have this functionality so that we can run workflows for release PRs. 

If it ends up taking too long in practice, we could perhaps modify this to run on the main branch so that it checks all the code snippets in the background after merging each PR.

PR #309, opened 2025-05-07
Add additional keywords to `mops.toml` for improved discoverability

Kamirus:
Adds mops keywords so that `new-base` would appear in the 'Data Structures' and 'Stable Memory' categories on https://mops.one/


rvanasa (approved):
Good idea; thanks!

PR #312, opened 2025-05-08
Fix `first` and `last` functions from the `List` module to return `null` on empty

### LINKED ISSUE #311: `List.last` has design issue
Traps on empty list although it returns ?T with trap "array out of bounds"
Having `null` instead of trap would be more convenient.
Also, the semantics seems currently different to `List.first`.
Another detail: The documentation could specify the empty case for `List.first`.

Kamirus:
- Make `first` and `last` return `null` on empty
- Expand on docs examples with similar functions returning nulls
- Cover in tests empty and emptied lists

Closes #311


test/List.test.mo
```diff
@@ -1220,4 +1276,52 @@ Test.suite(
...
+      "lastIndexOf",
+      func() {
+        Test.expect.bool(List.lastIndexOf(empty, Nat.equal, 0) == null).equal(true);
+        Test.expect.bool(List.lastIndexOf(emptied, Nat.equal, 0) == null).equal(true)
+      }
+    );
+    Test.test(
+      "firstIndexWhere",
+      func() {
+        Test.expect.bool(List.firstIndexWhere<Nat>(empty, func(x) = x == 0) == null).equal(true);
```
ggreif, line 1301: Why not
```suggestion
        Test.expect.bool(List.firstIndexWhere<Nat>(empty, func x = x == 0) == null).equal(true);
```
as well as below...

rvanasa (approved):
LGTM, thanks!

PR #315, opened 2025-05-16
Clarify maximum number of elements in List

rvanasa:
Addresses feedback from @timohanke.

### DIFF
--- a/src/List.mo
+++ b/src/List.mo
@@ -29,6 +29,8 @@ module {
   /// Since this is internally a two-dimensional array the access times for put and get operations
   /// will naturally be 2x slower than Buffer and Array. However, Array is not resizable and Buffer
   /// has `O(size)` memory waste.
+  ///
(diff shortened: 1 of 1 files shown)


src/List.mo
```diff
@@ -29,6 +29,8 @@ module {
   /// Since this is internally a two-dimensional array the access times for put and get operations
   /// will naturally be 2x slower than Buffer and Array. However, Array is not resizable and Buffer
   /// has `O(size)` memory waste.
+  ///
+  /// The maximum number of elements in a `List` is 2^32.
   public type List<T> = Types.List<T>;
```
AStepanov25, line 34: Actually the maximum number of elements is 2 ** 32 - 1, locate traps on 2 ** 32, I tried. It is because we store one redundant element in the beginning, because it helped to speed up indexing.

AStepanov25 (changes requested):

ggreif (approved):
LGTM!

PR #321, opened 2025-05-29
Add `findIndex()` function to data structures with `find()`

rvanasa:
Resolves #316.


AStepanov25:
In `List<T>` there are methods `firstIndexWhere` and `lastIndexWhere`. These methods can possibly be provided for `Array`, `Iter` so on.

Kamirus (approved):
LGTM

PR #329, opened 2025-06-12
Enable persistence of `Random` and `AsyncRandom` state in stable memory

rvanasa:
This PR adds a backwards-compatible way of persisting `Random` and `AsyncRandom` states in stable memory. It should also be possible to store the original `Random.seed()` and `Random.crypto()` as stable classes, making this approach preferable over using a module-based API as with data structures.

Examples:

```motoko
import Random "mo:core/Random";

persistent actor {
  // Original (works the same as before)
  transient let random = Random.seed(123);
  ignore random.natRange(0, 1000);

  // Alternative for stable memory
  let state1 = Random.seedState(123); // Initializes the state from a given seed
  transient let random1 = Random.seedFromState(state1);
  ignore random1.natRange(0, 1000);

  // Also available for system random to reuse entropy after canister upgrades
  // or provide custom entropy until resupply (e.g. for testing)
  let state2 = Random.cryptoState();
  transient let random2 = Random.cryptoFromState(state2);
  ignore random2.natRange(0, 1000);
}
```

The function names have room for improvement; suggestions welcome.

### DIFF
--- a/src/Random.mo
+++ b/src/Random.mo
@@ -32,8 +32,6 @@ module {
 
   public let blob : shared () -> async Blob = rawRand;
 
-
-
   /// Initializes a random number generator state. This is used
   /// to create a `Random` or `AsyncRandom` instance with a specific state.
   /// The state is empty, but it can be reused after upgrading the canister.


validation/api/api.lock.json
```diff
@@ -931,13 +931,17 @@
...
-      "public func fromAsyncGenerator(generator : () -> async* Blob) : AsyncRandom",
-      "public func fromGenerator(generator : () -> Blob) : Random",
-      "public class Random("
+      "public func crypto() : AsyncRandom<()>",
+      "public func cryptoFromState(state : State<()>) : AsyncRandom<()>",
+      "public func cryptoState() : State<()>",
+      "public func emptyState<T>(inner : T) : State<T>",
+      "public func fast(seed : Nat64) : Random<PRNG.State>",
+      "public func fastFromState(state : State<PRNG.State>) : Random<PRNG.State>",
+      "public class Random<T>(state : State<T>, generator : () -> Blob)",
```
Kamirus, line 942: This extra type parameter makes it more confusing.
The random classes come in two flavors: 'fast' or 'crypto', yes? Crypto is always async and Fast is always sync, yes?
How about removing the `<T>` from both AsyncRandom and Random classes? This type parameter is like an implementation detail.
How about this API:
```
"public class AsyncRandom(state : State, generator : () -> async* Blob)",
"public let blob : shared () -> async Blob",
"public func crypto() : AsyncRandom",
"public func cryptoFromState(state : State) : AsyncRandom",
"public func cryptoState() : State",
"public func fast(seed : Nat64) : Random",
"public func fastFromState(state : FastState) : Random",
"public type FastState",
"public class Random(state : State, generator : () -> Blob)",
"public func seedState(seed : Nat64) : FastState",
"public type State"
```

I'd make `emptyState` private and make an internal `StateT` type to hide the type parameter:
```motoko
public type State = {
  var bytes : [Nat8];
  var index : Nat;
  var bits : Nat8;
  var bitMask : Nat8
};

public type FastState = StateT<PRNG.State>;

type StateT<T> = State and {
  inner : T
};
```

Kamirus: It should be no breaking changes too, right?

validation/api/api.lock.json
```diff
@@ -931,13 +931,18 @@
...
+      "public class AsyncRandom(state : State, generator : () -> async* Blob)",
       "public let blob : shared () -> async Blob",
       "public func crypto() : AsyncRandom",
-      "public func fast(seed : Nat64) : Random",
-      "public func fromAsyncGenerator(generator : () -> async* Blob) : AsyncRandom",
-      "public func fromGenerator(generator : () -> Blob) : Random",
-      "public class Random("
+      "public func cryptoFromState(state : State) : AsyncRandom",
+      "public func cryptoState() : State",
+      "public func emptyState() : State",
```
Kamirus, line 939: When to use which of the two? I'd just remove one of them, maybe `cryptoState` because it is just equivalent to `emptyState`

Kamirus (approved):
LGTM!

PR #332, opened 2025-06-16
fix: Update readme to point to core

alexandru-uta:
Update readme file to point to mops `core` package instead of outdated `new-base`.


rvanasa:
Thanks for catching that!

PR #333, opened 2025-06-16
Breaking: rename `sliceToArray()` to `subArray()`

rvanasa:
Resolves #331.

This is a breaking change from previous versions of the new base / core library.

Alternative without the breaking change: https://github.com/dfinity/motoko-core/pull/377.


crusso (approved):

rvanasa:
Closed in favor of #377.

PR #334, opened 2025-06-18
feat: use the new VarArray tabulate primitive

christoph-dfinity:
This speed up various functions in VarArray

### DIFF
--- a/mops.toml
+++ b/mops.toml
@@ -17,7 +17,7 @@ test = "2.1.1"
 bench = "1.0.0"
 fuzz = "0.2.1"
 matchers = "https://github.com/kritzcreek/motoko-matchers#v1.3.1@5ba5f52bd9a5649dedf5e2a1ccd55d98ed7ff982"
-base-0-14-13 = "https://github.com/dfinity/motoko-base#moc-0.14.13@9e6010595cb9bad5b637ab7438fd11dc79a21a06"
+base-0-14-13 = "https://github.com/dfinity/motoko-base#moc-0.14.13@794174a307975c225cfb26b57f73e38a841c0415"
 
 [requirements]
 moc = "0.14.13"


src/VarArray.mo
```diff
@@ -62,19 +62,7 @@ module {
...
-    };
-    let newArray = Prim.Array_init<T>(size, array[0]);
-    var i = 0;
-    while (i < size) {
-      newArray[i] := array[i];
-      i += 1
-    };
-    newArray
-  };
+  public func clone<T>(array : [var T]) : [var T] = Prim.Array_tabulateVar<X>(array.size(), func i = array[i]);
```
Kamirus, line 65:
```suggestion
  public func clone<T>(array : [var T]) : [var T] = Prim.Array_tabulateVar<T>(array.size(), func i = array[i]);
```

christoph-dfinity:
@Kamirus I think I might need some help here. I need to bump the `motoko` npm dependency so it includes the new primitive, but we now also generate Wasm imports to the new `ic_root_key` system calls, which seems to not be supported by the replica used during docs validation (bit bizarre that that's a requirement).

Kamirus:
It uses dfx 0.27.0, maybe this one does not export the root key primitive yet? That's annoying :x 
Another reason for making these imports lazy 😄

Kamirus:
dfx 0.27.0 does not export [the root key](https://internetcomputer.org/docs/references/ic-interface-spec#0_42_0) yet...

Kamirus:
Solution coming here [#5276](https://github.com/dfinity/motoko/pull/5276)

christoph-dfinity:
Nice, looks like the update did it (another bump to the npm `motoko` package that is). I also added the requirements section as requested.

PTAL

mops.toml
```diff
@@ -17,7 +17,10 @@ test = "2.1.1"
 bench = "1.0.0"
 fuzz = "0.2.1"
 matchers = "https://github.com/kritzcreek/motoko-matchers#v1.3.1@5ba5f52bd9a5649dedf5e2a1ccd55d98ed7ff982"
-base-0-14-8 = "https://github.com/dfinity/motoko-base#moc-0.14.8@9e6010595cb9bad5b637ab7438fd11dc79a21a06"
+base-0-14-13 = "https://github.com/dfinity/motoko-base#moc-0.14.13@9e6010595cb9bad5b637ab7438fd11dc79a21a06"
```
Kamirus, line 20: The hash does not need to change? Anyway, we probably don't need to bump this dependency

Kamirus (approved):

PR #335, opened 2025-06-19
Setup `next-moc` branch

rvanasa:
Adds CI workflows to set up the `next-moc` branch in the same way as the previous base library repository.

Progress:
* [x] Documentation
* [x] `sync` CI workflow

Splitting into a future PR:
* Run tests using master branch of `motoko` repository for `next-moc`

### DIFF
--- a/Release.md
+++ b/Release.md
@@ -1,0 +1,23 @@
+# Release process
+
+Please refer to the [release instructions](https://github.com/dfinity/motoko/blob/master/Building.md#making-releases) in the Motoko repository.
+
+## The `next-moc` branch
+
+The `next-moc` branch contains changes that make base compatible with the
+in-development version of `moc`. This repository's public CI does _not_ run
+on that branch.
+
+External contributions are best made against `master`.
+
+- `master` branch is meant for the newest **released** version of `moc`
+  - The CI runs on this branch
+- `next-moc` branch is meant for the **in-development** version of `moc`
+  - This branch is used by the [`motoko` repository](https://github.com/dfinity/motoko)'s CI
+
+Both branches are kept in sync with each other by mutual, circular merges:
+- `next-moc` is updated automatically on each push to `master` via the [sync.yml](.github/workflows/sync.yml) workflow
+- `master` is updated **manually** on each release of `moc` as part of the `motoko` release process
+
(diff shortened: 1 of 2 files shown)


crusso:
Looks good! I guess we also need to actually create the next-moc branch (and then get motoko to pull from that).

Release.md
```diff
@@ -0,0 +1,23 @@
...
+  - The CI runs on this branch
+- `next-moc` branch is meant for the **in-development** version of `moc`
+  - This branch is used by the [`motoko` repository](https://github.com/dfinity/motoko)'s CI
+
+Both branches are kept in sync with each other by mutual, circular merges:
+- `next-moc` is updated automatically on each push to `master` via the [sync.yml](.github/workflows/sync.yml) workflow
+- `master` is updated **manually** on each release of `moc` as part of the `motoko` release process
+
+Only *normal* merges are allowed between `master` and `next-moc`, because development is permitted on both branches.
+This policy makes every PR (to either branch) visible in the history of both branches.
```
christoph-dfinity, line 23:
```suggestion
External contributions are best made against `main`.

- `main` branch is meant for the newest **released** version of `moc`
  - The CI runs on this branch
- `next-moc` branch is meant for the **in-development** version of `moc`
  - This branch is used by the [`motoko` repository](https://github.com/dfinity/motoko)'s CI

Both branches are kept in sync with each other by mutual, circular merges:
- `next-moc` is updated automatically on each push to `main` via the [sync.yml](.github/workflows/sync.yml) workflow
- `main` is updated **manually** on each release of `moc` as part of the `motoko` release process

Only *normal* merges are allowed between `main` and `next-moc`, because development is permitted on both branches.
This policy makes every PR (to either branch) visible in the history of both branches.
```

christoph-dfinity (changes requested):
Spotted a bit of `master` -> `main` churn in the Release.md instructions. Might be worth grepping for master in case I missed any occurrences.

Kamirus:
But do we actually need a separate branch now?
Having both `main` and `next-moc` has several drawbacks, like non-linear git history which results in horrible git history, need for automatic updates of both of them, general problems of having 2 main branches that are most of the time identical.
Can we avoid these for now? How frequent we are going to have incompatible changes between `motoko` and `motoko-core`? When was last time `next-moc` on the `motoko-base` repo was actually useful?

PR #336, opened 2025-06-23
Fix bug in List.last<T>.

AStepanov25:
There is a bug in List.last<T>: when elementIndex is zero we should look at the last element in the previous data block, not first. Probability to encounter this bug in 1 / sqrt(n). Optimized and cleaned up code. Added stronger tests.


AStepanov25:
I can't fix PR pipelines, there are 403 errors, maybe they don't work for external contributors?

Kamirus:
@rvanasa we should probably skip these 2 steps for external PRs, unless you have an idea how to fix them 😄

PR #337, opened 2025-06-24
Optimize methods in List.

AStepanov25:


src/List.mo
```diff
@@ -246,22 +284,37 @@ module {
...
-            case (null) null
+  public func map<T, R>(list : List<T>, f : T -> R) : List<R> {
+    let blocks = VarArray.repeat<[var ?R]>([var], list.blocks.size());
+    let blocksCount = list.blocks.size();
+
+    var i = 1;
+    while (i < blocksCount) {
+      let oldBlock = list.blocks[i];
+      let blockSize = oldBlock.size();
+      let newBlock = VarArray.repeat<?R>(null, blockSize);
```
timohanke, line 295: How would it perform if we used the VarArray.map function on the data blocks?

src/List.mo
```diff
@@ -1643,22 +1858,32 @@ module {
   ///
   /// *Runtime and space assumes that `compare` runs in O(1) time and space.
   public func compare<T>(list1 : List<T>, list2 : List<T>, compare : (T, T) -> Order.Order) : Order.Order {
-    let size1 = size(list1);
-    let size2 = size(list2);
-    let minSize = if (size1 < size2) { size1 } else { size2 };
+    let blocks1 = list1.blocks;
+    let blocks2 = list2.blocks;
+    let blockCount = Nat.min(blocks1.size(), blocks2.size());
```
timohanke, line 1863: Is this part worth it? The size calculation only happens once for the whole list. Cost does not depend on length of inputs.

src/List.mo
```diff
@@ -60,46 +60,54 @@ module {
   /// Runtime: `O(1)`
   ///
   /// Space: `O(1)`
-  public func singleton<T>(element : T) : List<T> = repeat(element, 1);
+  public func singleton<T>(element : T) : List<T> = {
+    var blockIndex = 2;
+    var blocks = [var [var], [var ?element]];
+    var elementIndex = 0
+  };
```
timohanke, line 67: Not sure if worth the optimization. Does not depend on length of list.

timohanke:
Not all improved functions are visible in the profiling table. For those, it is hard to tell from just the code diff to tell how much the improvement is and if it worth the inlining or not.

src/List.mo
AStepanov25: There can be initializations in cycle, if the data structure is `List<List<T>>` for example.

src/List.mo
```diff
@@ -549,12 +632,19 @@ module {
...
+    let (a, b) = do {
+      let i = Nat32.fromNat(index);
+      let lz = Nat32.bitcountLeadingZero(i);
+      let lz2 = lz >> 1;
+      if (lz & 1 == 0) {
+        (Nat32.toNat(((i << lz2) >> 16) ^ (0x10000 >> lz2)), Nat32.toNat(i & (0xFFFF >> lz2)))
+      } else {
+        (Nat32.toNat(((i << lz2) >> 15) ^ (0x18000 >> lz2)), Nat32.toNat(i & (0x7FFF >> lz2)))
+      }
+    };
```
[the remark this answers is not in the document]
AStepanov25: `put` and `getOpt` along with `get` are the most used methods of `List`, so they should be optimized at any cost. Anyway there a lot of code duplication there.

AStepanov25:
> Not all improved functions are visible in the profiling table. For those, it is hard to tell from just the code diff to tell how much the improvement is and if it worth the inlining or not.

I profiled only required functions, for others performance increase is obvious as the code similar to others functions implementations.

What functions are you interested in, I'll add profiling if needed.

src/List.mo
```diff
@@ -1643,22 +1858,32 @@ module {
...
-      switch (compare(next1(), next2())) {
-        case (#less) return #less;
-        case (#greater) return #greater;
-        case _ {}
+    var i = 1;
+    while (i < blockCount) {
+      let db1 = blocks1[i];
+      let db2 = blocks2[i];
+      let sz = Nat.min(db1.size(), db2.size());
+      if (sz == 0) return Nat.compare(size(list1), size(list2));
```
timohanke, line 1870: It seems that you are trying to delay the computation of `size(list1), size(list2)` as much as possible, and in most cases it never has to be calculated. Is that right? Is that because you think the calculation is expensive?

timohanke (approved):

Andrei1998:
We use the same operations in all benchmarks except number 4.) [which only uses `PriorityQueue.push`, all others do `PriorityQueue.pop` as well]:
- `PriorityQueue.push` uses `List.add`, `List.size`, `List.put`, `List.at`.
- `PriorityQueue.pop` uses `List.removeLast`, `List.isEmpty`, `List.size`, `List.put`, `List.get`, `List.at`.

Yet interestingly, the anomaly is for number 1.) [not number 4.)].

Number 1) basically consists of a random sequence of `PriorityQueue.push` and `PriorityQueue.pop`, with an equal probability of each entry being a `push` or a `pop`. The same anomaly does not show for 2), where `push` is twice as likely as `pop`. The length of the underlying `List` in 1.) and 2.) is a $\pm 1$ random walk, with the $+1$ and $-1$ probabilities being either $0.5-0.5$ or $0.66-0.33$. Number 3.) also does the same thing, just with an even higher probability to increase the length of the list. My current hypothesis is that what is special about 1.) is that it grows and shrinks the `List` many times, in contrast to the others, which mostly grow it, or grow and shrink it only once. Hence, this could point to something like a memory leak (maybe not exactly, but going in that direction).

timohanke:
The garbage increased by 0.4 bytes per operation. I wonder if the garbage increasing can actually mean that the code got better (for example freeing more after a pop)?

How often do you think the random walk hits length 0? I also wonder if it can be related to how much is freed when reaching the empty list again.

Andrei1998:
> The garbage increased by 0.4 bytes per operation. I wonder if the garbage increasing can actually mean that the code got better (for example freeing more after a pop)?

This could also be the case, indeed. It could be that we used to have extra data lying around for no good reason and we are now deallocating it more eagerly (more likely). Alternatively, it could be that we now also allocate more data, and hence also need to free more (less likely). Technically, if someone told us after a pop that the queue is gonna grow again (which in this benchmark should happen a lot), then it would be better to not deallocate memory eagerly, because the queue will grow again later. So, in general, it's a tradeoff: deallocate too eagerly and then reallocate later when the queue grows again, or deallocate less aggressively and risk no more operations coming in the future (and hence waste space long-term).

> How often do you think the random walk hits length 0? I also wonder if it can be related to how much is freed when reaching the empty list again.

Digging into this further might take more effort, but at least a [quick and dirty experiment](https://ideone.com/EVWoVW), written with the help of LLMs, shows that we should hit 0 around 500 times (the LLM computed around 504 when asked to do the math directly). Some of those 500 were probably already close to 0, but the expected maximum length of the list during the test should grow roughly with the square root of the number of operations. Of course, I could have just run the test with the fixed seed from the benchmark, but this should at least shed some preliminary light on the behavior.

PR #347, opened 2025-07-14
chore: fix boolean-flags in setup github action

### LINKED ISSUE #345: CI doesn't respect the `dfx` install flag's default
Defaults to `false`, but `dfx` still gets installed:

https://github.com/dfinity/motoko-core/actions/runs/16174651841/job/45702239899#step:4:190

Kamirus:
Fixes #345

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -5,13 +5,11 @@ inputs:
   setup-mops:
     description: Whether to setup mops
     required: false
-    default: true
-    type: boolean
+    default: "true"
   setup-dfx:
     description: Whether to setup dfx
     required: false
-    default: false
-    type: boolean
+    default: "false"
 runs:
   using: composite
   steps:
@@ -22,7 +20,7 @@ runs:
         cache: npm
 
     - name: Cache mops packages before installing dependencies
-      if: ${{ inputs.setup-mops }}
+      if: ${{ inputs.setup-mops == 'true' }}
       id: mops-cache
       uses: actions/cache@v4
       with:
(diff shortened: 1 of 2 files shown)


.github/actions/setup/action.yml
```diff
@@ -5,13 +5,11 @@ inputs:
   setup-mops:
     description: Whether to setup mops
     required: false
-    default: true
-    type: boolean
+    default: "true"
```
ggreif, line 8: Below you use single quotes `'` but here `"`. Why?

.github/actions/setup/action.yml
```diff
@@ -5,13 +5,11 @@ inputs:
   setup-mops:
     description: Whether to setup mops
     required: false
-    default: true
-    type: boolean
```
ggreif, line 9: So the `boolean` type is the culprit? Is this documented somewhere?

Kamirus: I don't know. It is just try and error 🤷‍♂️ I assume now that every input must be of string type

.github/actions/setup/action.yml
Kamirus: https://docs.github.com/en/actions/reference/evaluate-expressions-in-workflows-and-actions#literals
here is an example where `'` is used `${{ 'It''s open source!' }}`. `"` will throw an error

Kamirus: here I could probably use `'` too 
```suggestion
    default: 'true'
```
But I cannot omit `'`/`"`. 
If you want we can replace `"` with `'`.

ggreif (approved):
We'll see in Motoko nightly too. LGTM!

.github/actions/setup/action.yml
Kamirus: https://github.com/actions/runner/issues/1483

PR #355, opened 2025-07-23
updated Bool documentation to use new logicalAnd & logicalOr

P4P5T123:
Old function names logand(, ) and logor(, _) in Bool documentation [here](https://mops.one/core/docs/Bool)


rvanasa (approved):

rvanasa:
Thank you! Closing since we coincidentally happened to fix this in parallel as part of https://github.com/dfinity/motoko-core/pull/353.

PR #365, opened 2025-08-08
Breaking: fix `Int.fromText()` corner cases from original base library

### LINKED ISSUE #362: Edge case of function Int.fromText on input "-" and "+"
Int.fromText("-") and Int.fromText("+") both return '0' instead of 'null'. is this on purpose? 
Here's the branch of my current problem for easy reproduction of the bug: 
[PR](https://github.com/dfinity/motoko-training/pull/1349)

rvanasa:
Adjusts `Int.fromText()` to return `null` instead of `?0` for `"+"` and `"-"` (resolves #362).

This is a breaking change which would behave differently compared to the original base library. 

PR for including this in the migration guide: https://github.com/dfinity/motoko/pull/5388

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,7 @@
 # Next
 
 * Perf: Made `Nat.toText()` slightly more performant (#358).
+* Change `Int.fromText()` to return `null` for `"+"` and `"-"` instead of `?0` (#365).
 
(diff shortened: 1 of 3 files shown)


crusso (approved):
Nice!

PR #367, opened 2025-08-11
Breaking: return `?T` from `List.get()`

rvanasa:
Addresses feedback about inconsistent return types between `get()` functions across data structures.

Original usages of `List.get()` can be wrapped in `Option.unwrap()` for equivalent functionality.

### DIFF
--- a/src/List.mo
+++ b/src/List.mo
@@ -499,19 +499,46 @@ module {
     }
   };
 
-  /// Returns the element at index `index`. Indexing is zero-based.
-  /// Traps if `index >= size`, error message may not be descriptive.
+  /// Returns the element at index `index` as an option.
+  /// Returns `null` when `index >= size`. Indexing is zero-based.
   ///
   /// Example:
   /// ```motoko include=import
   /// let list = List.empty<Nat>();
   /// List.add(list, 10);
   /// List.add(list, 11);
-  /// assert List.get(list, 0) == 10;
+  /// assert List.get(list, 0) == ?10;
+  /// assert List.get(list, 2) == null;
   /// ```
   ///
   /// Runtime: `O(1)`
-  public func get<T>(list : List<T>, index : Nat) : T {
+  ///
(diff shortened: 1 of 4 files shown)


src/List.mo
```diff
@@ -499,43 +499,35 @@ module {
...
+    } else {
+      null
     }
   };
 
   /// Returns the element at index `index` as an option.
   /// Returns `null` when `index >= size`. Indexing is zero-based.
   ///
+  /// **Deprecated**: This function is now identical to `get()`. Use `get()` instead.
+  ///
```
Kamirus, line 530: Can we just rename `get -> at` and `getOpt -> get`? Without any other changes. I'd leave the inlined implementation in `at`, why change that now?
I think that would be a cleaner approach

crusso: I agree. So `at` for trapping, and `get` for option returning

src/pure/Set.mo
```diff
@@ -1126,9 +1127,9 @@ module {
       func buildFromSortedHelper(l : Nat, r : Nat, depth : Nat) : Tree<V> {
         if (l + 1 == r) {
           if (depth == maxDepth) {
-            return #red(#leaf, List.get(buf, l), #leaf)
+            return #red(#leaf, Option.unwrap(List.get(buf, l)), #leaf)
```
Kamirus, line 1130:
```suggestion
            return #red(#leaf, List.at(buf, l), #leaf)
```
and in other places

src/pure/Set.mo
```diff
@@ -39,6 +39,7 @@ import Iter "../Iter";
 import Types "../Types";
 import Nat "../Nat";
 import Order "../Order";
+import Option "../Option";
```
Kamirus, line 42:
```suggestion
```

Kamirus (changes requested):
Looks nice now!

PR #370, opened 2025-08-11
fix: Inconsistent handing of empty iterators in `Int.mo` and `Nat.mo`, and tests for `{Int, Nat}.rangeInclusive()`.

Andrei1998:
Range functions in `Int.mo` and `Nat.mo` have to return an empty iterator in some cases. Sometimes this is done explicitly (by returning an empty iterator), and sometimes implicitly (by relying on the underlying logic to resolve to an empty iterator). Most of the time, this is done using the former method, but there are a few instances where the latter is used (somewhat inconsistently):

- `Nat.range()` (but not `Nat*.range()`)
- `Nat.rangeBy()`
- `Nat.rangeInclusive()` (`Nat*.rangeInclusive()`)
- `Int.range()` (but not `Int*.range()`) 
- `Int.rangeInclusive()`(but not `Int*.rangeInclusive()`)

This PR fixes this inconsistency (applying the explicit approach to the above). This PR also adds more unit tests where appropriate, notably tests for `{Int, Nat}.rangeInclusive()`, which there weren't any so far.

This PR addresses https://github.com/dfinity/motoko-core/issues/369.

### DIFF
--- a/src/Int.mo
+++ b/src/Int.mo
@@ -499,7 +499,6 @@ module {
   /// ```
   public func rangeInclusive(from : Int, to : Int) : Iter.Iter<Int> {
     if (from > to) {
-      // Needs tests.
       Iter.empty()
     } else {
       object {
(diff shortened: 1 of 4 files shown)


alexandru-uta (approved):
Thanks, Andrei!

PR #372, opened 2025-08-12
feat: adds a Peakable iterator adapter that provides lookahead

christoph-dfinity:
I find myself using this a lot when writing Rust, but it might be too specific to compiler-y use-cases.

https://doc.rust-lang.org/beta/std/iter/struct.Peekable.html


christoph-dfinity:
CI is failing because I changed the `src/Types.mo` file, but the validate docs command fails on files with 0 snippets: https://github.com/dfinity/motoko-core/blob/a34a6aa5b6b1ad593744edbf09c3dcec3197d403/test/ts/validate/docs.ts#L148-L150

Not quite sure what to do about that. Is it really a problem if a file has 0 snippets?

rvanasa (commented):
Implementation LGTM, but perhaps too specialized for the `core` package? In the past, we suggested that these patterns should start as a Mops package which we could then absorb into core if we see a large number of downloads or frequent reimplementation.

christoph-dfinity:
Looks like there's an implementation of Peekable in the [`xtended-iter` package](https://mops.one/xtended-iter/docs/PeekableIter)

I'll close this PR for now.

PR #373, opened 2025-08-12
Add `List.at()` with the original behavior of `List.get()`

rvanasa:
See #367.

Naming precedent is the [`at()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at) method on JavaScript arrays.

In the future, we could expand this function to take `Int` values for indices relative to the end of the array (commonly used in Python and JS).

### DIFF
--- a/src/List.mo
+++ b/src/List.mo
@@ -523,6 +523,33 @@ module {
     }
   };
 
+  /// Returns the element at index `index`.
+  /// Traps if `index >= size`. Indexing is zero-based.
+  ///
+  /// Example:
+  /// ```motoko include=import
+  /// let list = List.empty<Nat>();
+  /// List.add(list, 10);
+  /// List.add(list, 11);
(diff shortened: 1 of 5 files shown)


src/List.mo
```diff
@@ -523,6 +523,33 @@ module {
...
+  /// List.add(list, 11);
+  /// assert List.at(list, 0) == 10;
+  /// // List.at(list, 2); // traps
+  /// ```
+  ///
+  /// Runtime: `O(1)`
+  ///
+  /// Space: `O(1)`
+  public func at<T>(list : List<T>, index : Nat) : T {
+    let (a, b) = locate(index);
```
crusso, line 542: I'd used the inlined implementation from the previous get.  Otherwise fine!

rvanasa:
Closing: combined with #367.

PR #374, opened 2025-08-12
Add doc comment / usage example to `Types` module

rvanasa:
Supersedes https://github.com/dfinity/motoko-core/pull/357.

### DIFF
--- a/src/Types.mo
+++ b/src/Types.mo
@@ -3,7 +3,7 @@
 /// Example usage:
 ///
 /// ```motoko name=import
-/// import { type Result, type Iter } "mo:core/Types";
+/// import { type Result; type Iter } "mo:core/Types";
 ///
 /// // Result for error handling
 /// let result : Result<Int, Text> = #ok(42);
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
         "fast-glob": "^3.3.2",
         "ic-mops": "^1.11.1",
         "mo-dev": "^0.13.0",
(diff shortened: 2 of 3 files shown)


src/Types.mo
```diff
@@ -1,3 +1,17 @@
+/// Common types used throughout the core package.
+///
+/// Example usage:
+///
+/// ```motoko name=import
+/// import Types "mo:core/Types";
```
christoph-dfinity, line 6:
```suggestion
/// import { type Result, type Iter } "mo:core/Types";
```

Could use the new type imports, which would maybe make this module a bit more useful, in that it just needs one import to add a bunch of unqualified types?

src/Types.mo
```diff
@@ -1,3 +1,17 @@
...
+///
+/// // Result for error handling
+/// let result : Types.Result<Int, Text> = #ok(42);
+///
+/// // Iterator for sequences
+/// let iter : Types.Iter<Nat> = { next = func() { ?1 } };
```
crusso, line 12:
```suggestion
/// let result : Result<Int, Text> = #ok(42);
///
/// // Iterator for sequences
/// let iter : Iter<Nat> = { next = func() { ?1 } };
```
(if you take Christophs suggestion above)

src/Types.mo
ggreif: The problem is that many users have a `moc` that doesn't feature this explicit import and could be surprised but this. Maybe doing this change a few months down the road would be more appropriate?

PR #388, opened 2025-09-02
feat: add weak reference library support

alexandru-uta:
This PR exports via the core library the Weak reference functionality.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,6 +7,7 @@
 
 * Add `sliceToVarArray()` to `Array` and `VarArray` (#377).
 * **Breaking:** Standardize function argument order (#376).
+* Add `binarySearch()` to `Array`, `VarArray`, and `List` modules (#375).
 * Add example usage documentation to `Types` module (#374).
 * Address inconsistent treatment of empty iterators by range functions in `Int` and `Nat` (#369).
 * Fix corner cases in `Nat.rangeByInclusive()` (#368).
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
         "motoko": "^3.14.3",
         "npm-run-all": "^4.1.5",
         "prettier": "2",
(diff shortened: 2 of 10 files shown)


src/WeakReference.mo
```diff
@@ -0,0 +1,30 @@
...
+module {
+  type WeakReference<T> = {
+    ref : weak T
+  };
+
+  public func allocateWeakRef<T>(obj : T) : WeakReference<T> {
+    return { ref = Prim.allocWeakRef<T>(obj) }
+  };
+
+  public func deref<T>(weakRef : WeakReference<T>) : ?T {
```
rvanasa, line 22: `WeakReference.get`? This would be more consistent with other languages such as Java and C#.

src/WeakReference.mo
```diff
@@ -0,0 +1,30 @@
...
+    return Prim.weakGet<T>(weakRef.ref)
```
rvanasa, line 23: Do we need the `WeakReference<T>` type, or would it be possible to simply pass `weak T` to the function (similar to `WeakReference.isLive`)? We probably want the input type to be the same for this function and `isLive`.

src/WeakReference.mo
```diff
@@ -0,0 +1,30 @@
...
+/// ```
+
+import Prim "mo:⛔"
+
+module {
+  type WeakReference<T> = {
+    ref : weak T
+  };
+
+  public func allocateWeakRef<T>(obj : T) : WeakReference<T> {
```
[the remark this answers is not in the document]
alexandru-uta: makes sense, thanks! I changed to `allocate`

src/WeakReference.mo
alexandru-uta: you're right, they need to be the same -- but I ended up modifying the other one; since we promote using this module, then I guess we need to use it consistently, otherwise there's no need to hide the `weak T`.

rvanasa (approved):

PR #390, opened 2025-09-08
Persistent object-oriented data structures

luc-blaeser:
Based on https://github.com/dfinity/motoko/pull/5451
* Persistent classes for `List`, `Map`, `Set`, and `Stack`.
* Fully backwards compatible.
* Persistent `compare` functions for primitive types.

TODO:
* Unit tests
* Documentation
* CI needs to use new compiler with persistent functions/classes


rvanasa (commented):
Thank you for preparing this PR! 

What do you think of using a more concise import pattern, e.g. `mo:core/class/List` or `mo:core/oo/List` in place of `mo:core/object-oriented/List`?

PR #397, opened 2025-09-27
experiment: use implicits on Map

christoph-dfinity:


src/Map.mo
```diff
@@ -1260,11 +1260,11 @@ module {
   /// assuming that `keyFormat` and `valueFormat` have runtime and space costs of `O(1)`.
   ///
   /// Note: Creates `O(log(n))` temporary objects that will be collected as garbage.
-  public func toText<K, V>(map : Map<K, V>, keyFormat : K -> Text, valueFormat : V -> Text) : Text {
+  public func toText<K, V>(map : Map<K, V>, toText : (implicit : K -> Text), valueFormat : V -> Text) : Text {
```
crusso, line 1263:
```suggestion
  public let toText : <K, V>(map : Map<K, V>, toText : (implicit : K -> Text), toText : V -> Text) : Text =
       func <K, V>(map : Map<K, V>, toText : K -> Text, valueFormat : V -> Text) : Text {
```
Might just work (even now)

crusso: No longer necessary since we can just override the default implicit name

PR #401, opened 2025-10-02
feat: Add `isCleanReject` to the `Error` module, align reject code order with spec and improve comments

Andrei1998:
Add `isCleanReject(error : Error) : Bool` as proposed in https://github.com/dfinity/motoko-core/issues/400.

Side changes:
- Update the order of `ErrorCode` variants corresponding to reject codes to match the [IC interface specification](https://internetcomputer.org/docs/references/ic-interface-spec#reject-codes).
- Improve comments for `#canister_error` and `#system_unknown`.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,6 @@
 ## Next
 
-* Add `isCleanReject` to `Error` and align reject code order with IC interface specification (#401).
+* Add `isCleanReject` to `Error`, align reject code order with IC interface specification and improve comments (#401).
 * internal: updates `matchers` dev-dependency (#394).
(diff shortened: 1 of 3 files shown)


Andrei1998:
I couldn't think of a good snippet to put in the docstring, hence suggestions are welcome.
My take would be to add it to the snippet in the docstring of `isRetryPossible` below, and I am thinking to do this in a follow-up PR where I revamp the docstring for `isRetryPossible` (part of the other proposal in https://github.com/dfinity/motoko-core/issues/400).

rvanasa (approved):
LGTM; thanks!

PR #408, opened 2025-10-19
Added isSorted to List.

AStepanov25:


rvanasa (approved):
LGTM, and it seems reasonable for us to add `isSorted()` to all other data structures with `sort()` / `sortInPlace()` (I'll do this in a separate PR).

PR #409, opened 2025-10-19
Add truncate to List.

AStepanov25:


timohanke:
Is the trapping behaviour when out of bound desired?

It could be more convenient for the code using this if truncating out of bound would do nothing. I mean in the same sense as a half-open interval [0,-1) is in some languages silently accepted as an empty interval and doesn't trap.

Any opinions?

timohanke:
LGTM to merge. Thanks!

rvanasa (approved):

PR #411, opened 2025-10-19
Add forEachInRange to List.

AStepanov25:


timohanke:
There are probably many functions in the List module that could benefit from a variant that operates on a "range" (= sublist) instead of on the whole list. This is just one example. Therefore it might make sense to hold off on this one and wait to see all the other functions. Then it will be easier to find a consistent naming scheme and consistent argument pattern to define the range.

The current name forEachRange isn't consistent with forEachEntry because we are not applying something "for each range". Maybe `forEachInRange`?

Regarding argument pattern:
* shall it be `list, from, to, f` instead of `list, f, from, to`?
* shall `list, from, to` be grouped into a tuple?

We can also wait for implicits to be released because range start or end could have a default value.

In light of this, shall we make this PR a draft at this point so it doesn't accidentally get merged?

timohanke:
Ok, let's rename to `forEachInRange` and keep arguments order as-is.

PR #420, opened 2025-11-03
experiment: conditionally deprecate types and values for caffeine using #M0235

crusso:


crusso:
@m-styp in the parent on master we added a bunch of toIntXXX/fromIntXXX and toNatXXX/fromNatXXX conversions, and I was trying to deprecate the new fromIntXXX and fromNatXXX conversions in this branch. However, I noticed that only fromNatXXX conversions are deprecate, but all the fromIntXXX conversion are not deprecated. Is that intentional?

Anyway, in the end I only did https://github.com/dfinity/motoko-core/pull/420/commits/51f82bd30832d4ff13dd8a59ea0432e4a0eaf776 (deprecating only the new fromNatXXX conversions).

PR #423, opened 2025-11-10
Add reader function to List.

AStepanov25:
Unsafe iterator starting with some position. This is a PR to open a discussion, docs and tests will follow.


timohanke:
Thanks!

Example: This will be very handy if people want to hash from a `List<Nat8>`. Then they can just do it efficiently with functions that are already provided by the new Sha2 package developed for motoko 1.0. Without this function I would have to build special List support into the Sha2 package which is kind of ugly.

src/List.mo
```diff
@@ -2891,4 +2891,47 @@ module {
...
+        db := list.blocks[blockIndex];
+        dbSize := db.size();
+        elementIndex := 0
+      };
+      switch (db[elementIndex]) {
+        case (?ret) {
+          elementIndex += 1;
+          return ret
+        };
+        case (_) Prim.trap(INTERNAL_ERROR)
```
rvanasa: Maybe a different trap message to indicate that this an out-of-bounds access rather than an internal error:

```suggestion
        case (_) Prim.trap("List.reader(): out of bounds")
```

Note that `List.reader()` with the parentheses is a precedent set by other modules for the trap message.

PR #428, opened 2025-11-19
Added fillWith to List.

AStepanov25:


timohanke:
I think I would skip this function (not merge it).

PR #431, opened 2025-11-25
Add full conversion matrix for `Nat*` types

rvanasa:
Expands on #430 by adding conversions between any two `Nat` types.


alexandru-uta (approved):
nice, this would be very helpful!

PR #486, opened 2026-04-12
Optimize `Array.fromIter`

timohanke:
Makes it 20% faster.


ggreif (approved):
Good changes!

timohanke:
> Good changes!

Did the same for `VarArray.fromIter` then.

Background why to optimise this particular function? Because it arises when migrating `base/Buffer` to `core`. It is one of the possible migrations, applicable to certain uses of `Buffer`.

PR #487, opened 2026-04-14
Add benchmarks for all modules

rvanasa:
Adds extensive benchmarking for all modules to monitor performance across new compiler versions and changes to the core package implementation.


timohanke:
If you are doing mass creation of benchmarks then you might want to consider moving away from the bench package (which passes row/col information to the runner as Text) as use an index based approach. The reason is that the Text matching becomes part of the measurement and distorts relative comparisons for any short running benches. For example, looking at `modules/Array.bench.mo` the runner contains this kind of switch statement:
```
        let arr = switch col {
          case "100" a100;
```
and as well
```
        switch row {
          case "all" ignore Array.all<Nat>(arr, func x = x < 2_000_000);
```
The Text comparisons are expensive. For a line like this in the output table:
```
| any          |   1_982 |     3_134 |      4_280 |
```
that is probably the majority of the instructions.

The most precise approach I have found is the one taken in this example:  https://github.com/research-ag/bench-helper/blob/main/bench/04-dynamic-routines.bench.mo
This one avoids Text comparison and also avoids switch as well (because with switch "later" tests become more expensive than "earlier" ones). There are other approaches as well in the same directory which don't use Text (but they may use switch).

AI is quote good at mass converting benchmarks to this approach. I have used this skill file for it in the past: https://github.com/research-ag/motoko-skills/blob/main/skills/benchmarks-generation/SKILL.md

PR #489, opened 2026-04-15
refactor: use contextual-dot throughout

christoph-dfinity:
Migrates tests and related tooling to contextual receiver syntax with implicit compare/equality where applicable.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,7 @@
 ## Next
 
+* Bump the `motoko` npm package used for doc-snippet checks, fix iterator call patterns (`entries`/`values`), and correct a few documentation examples (#489).
+
 ## 2.2.0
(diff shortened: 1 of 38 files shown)


christoph-dfinity:
Hold off on reviewing, I'll do a proper review myself first.

Kamirus:
Maybe ask Cursor to enable the context dot warning as error, we can add that to the mops.toml `[moc].args` and enforce that `mops check src/**/*.mo` passes in CI (just an idea) and also to the doc verifier (so that we know all examples in the docstrings use context dot where possible).
You can just forward this message to Cursor.
btw mops skill: https://github.com/caffeinelabs/mops/blob/main/.agents/skills/mops-cli/SKILL.md

PR #491, opened 2026-04-20
feat: adds accessors for caller attributes

christoph-dfinity:
Requires https://github.com/caffeinelabs/motoko/pull/5970

Fixes LANG-1250

### DIFF
--- a/src/CallerAttributes.mo
+++ b/src/CallerAttributes.mo
@@ -2,22 +2,12 @@
 /// TODO: link to official documentation, once it's available.
 
 import Prim "mo:⛔";
-import Test "Text";
+import Text "Text";
 import Iter "Iter";
 import Runtime "Runtime";
 import Principal "Principal";
 
 module {
-
-  func getSigner<system>() : ?Principal {
-    let signer : Blob = Prim.callerInfoSigner<system>();
-    // An empty signer means no attributes where sent in this call.
-    if (signer.size() == 0) {
-      return null
-    };
-    ?Prim.principalOfBlob(signer)
-  };
-
   /// Returns the attribute data attached to the current call, but only
   /// when the signer is listed in the `trusted_attribute_signers`
   /// canister environment variable.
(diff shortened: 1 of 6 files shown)


src/CallerAttributes.mo
```diff
@@ -0,0 +1,80 @@
...
+  ///       case (?data) { /* attributes came from a trusted signer */ };
+  ///       case null { /* no attributes, or signer is not trusted */ };
+  ///     };
+  ///   };
+  /// }
+  /// ```
+  public func getAttributes<system>() : ?Blob {
+    let ?signer = getSigner<system>() else { return null };
+    let ?trustedSigners = Runtime.envVar<system>("trusted_attribute_signers") else {
+      return null
```
aterga, line 41: I would recommend trapping in this case (the caller should not attempt to send attributes signed by someone the consumer doesn't recognize).

src/CallerAttributes.mo
```diff
@@ -0,0 +1,80 @@
...
+    };
+    let trustedSignerList = trustedSigners.split(#char(','));
+
+    if (trustedSignerList.any(func(t) { Principal.fromText(t) == signer })) {
+      return ?Prim.callerInfoData<system>()
+    };
+    return null
```
aterga: [Ditto](https://github.com/caffeinelabs/motoko-core/pull/491/changes#r3110829477)

src/CallerAttributes.mo
```diff
@@ -0,0 +1,80 @@
...
+  /// persistent actor {
+  ///   public shared func handle() : async () {
+  ///     switch (CallerAttributes.getUntrustedAttributes()) {
+  ///       case (?(signer, data)) { /* inspect signer and data */ };
+  ///       case null { /* no attributes on this call */ };
+  ///     };
+  ///   };
+  /// }
+  /// ```
+  public func getUntrustedAttributes<system>() : ?(Principal, Blob) {
```
aterga: Why do we need this function?

christoph-dfinity: To allow the more involved auth schemes Thomas requested. (per method/per attribute)

sea-snake: It enables custom authorization schemes indeed, but if one can write the following in such case (e.g. in a SDK):
```
let ?signer = getSigner<system>() else { return null };
let attributes = Prim.callerInfoData<system>();
```

Then this method isn't strictly needed and could potentially, rather be a risk factor for devs/AI accidentally using this method if there's an error with the trusted method.

christoph-dfinity: User code should not require `Prim`, so we'd need to cover the functionality from within core.

I've removed the untrusted function for now, it's easy to add later on when we know a bit more about the requirements.

alexandru-uta (approved):

PR #497, opened 2026-05-07
feat: add `Array.toBlob` and `VarArray.toBlob`

Kamirus:
Adds `Array.toBlob` and `VarArray.toBlob` so byte arrays can be converted via dot notation, mirroring `Principal.toBlob`:

```motoko
let blob = bytes.toBlob();   // was: Blob.fromArray(bytes)
```

Deprecates the now-redundant `Blob.fromArray`, `Blob.fromVarArray`, and `VarArray.fromIter` (M0235), matching the existing pattern for `Iter.fromArray`, `Array.fromVarArray`, etc.


christoph-dfinity (approved):
LGTM

PR #499, opened 2026-05-15
chore: bump `moc` to 1.8.0

ggreif:


mops.toml
```diff
@@ -21,8 +21,8 @@ base-0-14-13 = "https://github.com/dfinity/motoko-base#moc-0.14.13@794174a307975
 bench-helper = "0.0.3"
 
 [requirements]
-moc = "1.6.0"
+moc = "1.8.0"
```
Kamirus: no need to bump requirements, 1.6.0 is still compatible with this core version, yes?

Kamirus (changes requested):

mops.toml
ggreif:
```suggestion
moc = "1.6.0"
```

Kamirus (approved):

PR #500, opened 2026-06-02
chore: enforce M0223 (redundant type instantiation) as error

Kamirus:
Promote `M0223` (redundant type instantiation) from off-by-default to a hard error, in both source and doc snippets, and strip the existing violations so the wall stands.

- `mops.toml` → `[moc] args = ["-E=M0223"]` (covers all `.mo` source).
- `test/ts/validate/docs.ts` → adds the same flag so `///` snippets that ship in our public docs can't drift either.
- Mechanical autofix across source + tests + ~190 doc snippets (no behavior change).

```mo
-let array = Prim.Array_init<T>(size, self[0]);
+let array = Prim.Array_init(size, self[0]);
```

### DIFF
--- a/mops.toml
+++ b/mops.toml
@@ -26,3 +26,6 @@ moc = "1.6.0"
 [toolchain]
 moc = "1.8.0"
 wasmtime = "44.0.0"
+
+[moc]
+args = ["-W=M0223"]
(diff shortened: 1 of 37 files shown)


christoph-dfinity (approved):
LGTM

PR #501, opened 2026-06-04
Deprecate `fromX` conversions with hints

Kamirus:
`Module.fromX(x)` conversions that have a `Module.toX` counterpart now fire a standard `@deprecated` warning by default and name the replacement:

```
warning [M0154], field fromArray is deprecated
note: Use `Array.toBlob` instead.
```

Previously these used the caffeine-only `M0235` lint — silent for everyone except caffeine users, and the message was just `"is deprecated for caffeine"` with no migration path. Hints always name a fully-qualified `Module.toX` function — easier to follow than a `bytes.toBlob()` snippet that relies on a particular variable name and the right module being imported.

**Before / after**

```motoko
// Before — warning was caffeine-only, no hint
let blob = Blob.fromArray(bytes);

// After
//   warning [M0154], field fromArray is deprecated
//   note: Use `Array.toBlob` instead.
let blob = Array.toBlob(bytes);
```

**Newly deprecated (had no deprecation before)**

`Nat.fromNat8/16/32/64`, `NatN.fromNat`, `Float.fromFloat32`, `Float32.fromFloat`, `Text.fromChar`, `Char.fromNat32`, plus the `Int.fromX` / `IntN.fromIntM` family (which was skipped in [#420](https://github.com/dfinity/motoko-core/pull/420) — applying the same rule for consistency).

**What is unchanged**

`Map.insert`/`delete`, `Set.insert`/`delete`, `assertValid`, `toPure`/`fromPure`, the stateful `Random` class, and the pure type aliases stay `M0235`-deprecated — those are caffeine-specific preferences, not universal "prefer `toX`" rules.

**Docs guard**

`validate:docs` now treats `@deprecated` (`M0154`) usages as errors so documentation examples can never reference a deprecated API (`M0235` remains a warning, since it's caffeine-specific).

Follow-up to [#497](https://github.com/dfinity/motoko-core/pull/497). The un-deprecation of `List.get`, `Map.{swap,replace,take}`, `Set.{deleteAll,insertAll}`, and `Result.Result` was split into [#502](https://github.com/dfinity/motoko-core/pull/502) and is already merged.


alexandru-uta (approved):
thanks!

PR #508, opened 2026-06-29
chore: bump toolchain moc → 1.11.0

ggreif:
Bumps `[toolchain] moc` from 1.8.0 to **1.11.0** (just released) so CI exercises motoko-core against the new compiler. Pure verification/adoption of the new toolchain; `[requirements] moc` (1.6.0 minimum) unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/mops.toml
+++ b/mops.toml
@@ -24,7 +24,7 @@ bench-helper = "0.0.3"
 moc = "1.6.0"
 
 [toolchain]
-moc = "1.8.0"
+moc = "1.11.0"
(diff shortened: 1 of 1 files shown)


ggreif:
Not merging — keeping core's toolchain pinned to moc 1.8.0. The library shouldn't force consumers onto a newer moc; as long as it builds against the pinned 1.8.0 (which CI verifies), that's the contract. This PR was just a one-off local check against the freshly released 1.11.0.

PR #512, opened 2026-07-14
fix: correct List range/slice/get bounds handling

christoph-dfinity:
- range: clamp negative `fromInclusive` so the iterator stops at
  `toExclusive` instead of yielding elements past the requested window
- sliceToArray/sliceToVarArray: clamp `end` to `start` in actualInterval
  so an inverted range returns an empty slice instead of trapping on
  `end - start` underflow
- get: return null for `index >= 2^32` instead of trapping in Nat.toNat32

Add regression tests covering negative and inverted index arguments.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,4 +1,5 @@
 ## Next
+*  fix: correct List range/slice/get bounds handling (#512).
 
 ## 2.6.0
 * Add `Base64.decode : Text -> ?Blob`, the inverse of `Base64.encode` (RFC 4648 standard alphabet) (#507).
--- a/src/List.mo
+++ b/src/List.mo
@@ -1100,6 +1100,7 @@ module {
   ///
   /// Space: `O(1)`
   public func get<T>(self : List<T>, index : Nat) : ?T {
+    if (index >= 0x1_0000_0000) return null;
     // inlined version of locate
     let (a, b) = do {
       let i = Nat.toNat32(index);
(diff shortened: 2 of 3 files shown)


alexandru-uta (approved):
TY

timohanke:
Not sure how much we still care about 32-bit builds. But the new line 
```
if (index >= 0x1_0000_0000) return null;
```
introduced in `get` is problematic in 32-bit builds. It increases the instruction cost of `get` by 3.5x _and_ creates heap allocations whereas before `get` was allocation-free. For 64-bit builds the increase in instructions in small and `get` remains allocation-free.

Better code would be:
```
if (Prim.rightShift(index, 32) != 0) return null;
```

Moreover, with this new line in place, three lines below we can replace
```
let i = Nat.toNat32(index);
```
with the faster
```
let i = Nat32.fromIntWrap(index);
```

See #515  for a fix. The PR also fixes a bug in List.size() for List size 2^32 along the way. And removes unnecessary allocations in 3 other functions besides get().

timohanke:
Question regarding negative fromInclusive/toInclusive in range/slice. Do we really want to allow the case with fromInclusive negative and toExclusive positive as in `List.range(vec, -3, 3)`. What would be the application? I can understand the other cases but not this one. It seems confusing.

PR #513, opened 2026-07-14
fix: restore pure/Queue invariant on popFront/popBack

christoph-dfinity:
## Problem

`pure/Queue` is the two-list amortized deque `(front, size, backReversed)`. Its accessors `peekFront`, `peekBack`, and `contains` rely on the invariant that *if one access list is empty, the queue has ≤1 element*. `push`/`check` maintain it, but `popFront`/`popBack` returned the reduced queue **without** re-balancing:

```motoko
case (?(i, f), n, b) ?(i, (f, n - 1, b));   // popFront
case (f, n, ?(i, b)) ?((f, n - 1, b), i);   // popBack
```

When the queue is lopsided (a normal state after repeated `pushBack`/`pushFront`), popping the last element of one access list left the other list with >1 element. A *subsequent* pop self-heals via the `case _ → popFront(check self)` fallback, so pop-sequences worked — but the accessors observed the bad state directly:

- **Debug/test builds:** `debug assert` trips → **trap**.
- **Release builds:** silently returns the **wrong answer** (e.g. `peekFront` returns `null` on a 3-element queue).

`values`/`toArray` were unaffected (they read both lists), so only the O(1) accessors were hit.

## Fix

Route the reduced queue through `check` in both pop cases. `check` short-circuits (`case q q`) when both lists are non-empty — the common case — so this is O(1) there and amortized cost is unchanged; the split only runs when a pop actually empties one side.

## Tests

Added a `"pop restores the queue invariant"` suite covering both directions (back-loaded + `popFront`, front-loaded + `popBack`), asserting `peekFront`/`peekBack`/`size`/`contains`/`toArray` after the pop.

Verified with the `moc` interpreter against the working tree: the new suite **traps on the old code** and **passes on the fixed code**; all four queue suites (`pure/Queue`, `Queue`, `PriorityQueue`, `pure/RealTimeQueue`) remain green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,5 @@
 ## Next
-* Fix `pure/Queue.peekFront`, `peekBack`, and `contains` returning wrong results (and trapping in debug builds) after a `popFront`/`popBack` empties one of the internal access lists. `popFront`/`popBack` now rebalance their result so the queue invariant is always restored.
+* Fix `Queue` and `pure/Queue` for certain sequences of push + peek (#513).
 
 ## 2.6.0
 * Add `Base64.decode : Text -> ?Blob`, the inverse of `Base64.encode` (RFC 4648 standard alphabet) (#507).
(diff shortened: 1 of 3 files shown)


alexandru-uta (approved):
ty

PR #515, opened 2026-07-19
Ensure zero-allocation List functions; drop List size limit

timohanke:
This PR does two things:

1.)
Adds an allocation-regression test (wasi testmode) that measures Prim.rts_total_allocation() around each `List` function that is allocation-free today and asserts a zero delta, so later commits cannot accidentally introduce allocations on hot paths.

Covers 34 functions. Makes the following allocation-free:
- get (had allocation under legacy persistence)
- find
- equal (had allocation per element)
- compare (had allocation per element)

2.)
Drops the prior List size limit.
Background: Prior to this PR, List behaviour was only correct up to a size of 2^32 - 1. Growing the List past that was possible but certain bugs started to occur at exactly 2^32. This PR first fixes all bugs that occur at exactly 2^32:

- size, binarySearch, lastIndexOf, prevIndexOf, truncate, addRepeat : fix bugs that occur if the List's size or the function's argument or return value reaches `2^32`

But after this is done the List starts to work up to size 2^61.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -113,7 +113,6 @@ jobs:
       - uses: ./.github/actions/setup
       - run: npm test
 
-      # Requires mops >= 2.17.0 for the `-- <moc flags>` passthrough
       - name: Tests under legacy persistence
         run: npx ic-mops test List.allocation -- --legacy-persistence
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -113,6 +113,12 @@ jobs:
       - uses: ./.github/actions/setup
       - run: npm test
 
+      # Rewrite the [moc] args line in place
+      - name: Tests under legacy persistence
+        run: | 
+          perl -pi -e 's/^args = .*/args = ["--legacy-persistence"]/' mops.toml
+          npx ic-mops test List.allocation
+
   bench:
     needs: check-motoko-changes
     if: |
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -113,11 +113,9 @@ jobs:
       - uses: ./.github/actions/setup
       - run: npm test
 
-      # Rewrite the [moc] args line in place
+      # Requires mops >= 2.17.0 for the `-- <moc flags>` passthrough
       - name: Tests under legacy persistence
-        run: | 
-          perl -pi -e 's/^args = .*/args = ["--legacy-persistence"]/' mops.toml
-          npx ic-mops test List.allocation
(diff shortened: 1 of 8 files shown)


christoph-dfinity (approved):
Thank you for taking the time and spotting the perf regression. I'm fine with merging the "hack" in tests.yml for now and removing it again once we've merged and released https://github.com/caffeinelabs/mops/pull/615

timohanke:
I see. There is https://github.com/caffeinelabs/mops/pull/616 with the feature. Cool!

christoph-dfinity:
Yeah, sorry I linked the wrong PR number. That has since been released as 2.17.0, would you like to do the bump in this PR, or should I do it against `main` and we rebase/merge this PR on top?

timohanke:
I made the bump here.

.github/workflows/tests.yml
```diff
@@ -113,6 +113,10 @@ jobs:
       - uses: ./.github/actions/setup
       - run: npm test
 
+      # Requires mops >= 2.17.0 for the `-- <moc flags>` passthrough
```
christoph-dfinity: No need for this comment, we're not going to go backwards with mops versions.

christoph-dfinity (approved):
Thanks! Just one minor nit

timohanke:
Update:
Fixing all bugs that occur at the exact size of 2^32 required changing some Nat32-based arithmetic to Nat64. (Didn't have performance impact.) It also required adding a guard into `add` and other growing functions to prevent the List from growing beyond 2^32.

But with the Nat64 arithmetic in place the List can now actually support larger sizes. The guard limiting growth to 2^32 is completely arbitrary. So I dropped it now. It was a trivial commit, all the work had already been done as part of the bug fixes.

The new size limit is exactly 2^61. The limiting factor is now the arithmetic used for datablock indices which is still Nat32. I kept that in place (did not change to Nat64) to keep the PR smaller. Raising the limit to somewhere between 2^61 and 2^64 is irrelevant anyway.

One question I have about `get`. In PR #512 we made sure that `get` returns null for index arguments beyond the List's size. The question is if that is really necessary for Nat arguments beyond 2^64. Or if we can make it such that `get` behaviour is:
- trap for index >= 2^64
- return null for index in [list.size(), 2^64]
- return ?x for index < list.size() 

That would allow us to drop a guard in `get` that in practice never fires.

Kamirus:
> * trap for index >= 2^64

I think the whole idea with this guard was to avoid trapping, because we prioritize this extra "safety" over performance.
By "safety" I mean the guarantee that if this function returns a nullable value then it should never trap.
Anyway, let's ship the PR before it gets stale

